PR #8, opened 2023-05-12
Define healthCheck params to avoid force replace

lorenyu:
## Ticket

{LINK TO TICKET}

## Changes
see title

## Context for reviewers
We had some stuff in the service module that would force a replacement of the task definition regardless of whether anything changed. This change fixes that by adding some values in the container definition that AWS was setting default values for.

`terraform plan` in `infra/app/envs/dev` before change (on `main`):
image

After change:
image

Note that this PR is for demonstration since this test repo has an actual environment. The actual change will be made in template-infra.

## Testing
see context section


lorenyu:
done as part of https://github.com/navapbc/template-infra/pull/265

PR #12, opened 2023-05-18
Fix CI

lorenyu:
## Changes
* Remove extra call to terraform init that doesn't pass backend-config
* Update terraform init call in build-repository to pass backend-config

## Context for reviewers
Didn't realize that infra e2e tests were broken since they were commented out.
This change fixes the e2e tests. It passes in backend-config to the terraform init calls.

## Testing
Ran `make infra-test` locally, and will run it in CI (e2e tests job) on this PR too


lorenyu:
done as part of https://github.com/navapbc/template-infra/pull/274

PR #16, opened 2023-06-15
Support github action workflows for apps with multiple AWS accounts

lorenyu:
## Ticket

Work for https://github.com/navapbc/template-infra/issues/314

## Changes
* Add custom GitHub action configure-aws-credentials that wraps aws-actions/configure-aws-credentials.
* Add account_names_by_environment configuration to app-config.
* Update github action workflows to use the local custom action

Misc changes
* Add help comments to a couple of Makefile commands

## Context
Whenever we have a GitHub workflow that needs to access AWS resources (e.g. through terraform), we need to have a "Configure AWS credentials" using aws-actions/configure-aws-credentials and pass in the IAM role that the GitHub Actions runner should assume. Currently this role is hardcoded directly into the workflow files themselves.

There are a few problems with this:

1. Most importantly, this won't work with projects that have multiple AWS accounts (e.g. one account per environment), since the role ARN will be different deploying to different environments
2. Projects need to manually update the role-to-assume when setting up their project

These changes add a new local custom GitHub action configure-aws-credentials that wraps aws-actions/configure-aws-credentials. The custom action takes an app name and environment name, then uses a new application configuration map account_names_by_environment that maps environments to AWS accounts to determine which AWS account to authenticate into. The account_names_by_environment application configuration and can be customized to define a multi-account setup, a single-account setup, or something else.

## Testing

- [X] Run build-and-publish.yml from this branch
   https://github.com/navapbc/platform-test/actions/runs/5740584945
   
   image

- [x] Run ci-infra.yml from this branch (this will happen automatically as part of CI)
- [x] Run cd.yml from this branch
   https://github.com/navapbc/platform-test/actions/runs/5742072041


lorenyu:
as part of this PR, may want to update the new instructions in set-up-app-env.md that https://github.com/navapbc/template-infra/pull/364/files is adding

lorenyu:
Done as part of https://github.com/navapbc/template-infra/pull/319

PR #17, opened 2023-06-16
Move database params from variables file to config module

lorenyu:
## Ticket

Relates to https://github.com/navapbc/template-infra/issues/301

## Changes
* Add env-config submodule of app-config
* Add environment configs to app-config
* Use env config rather than variable to construct db vars
* Remove now unneeded db_vars variable from infra/app/service module
* Update service to accept list of db security group ids

## Context for reviewers
@acouch [pointed out in this PR comment](https://github.com/navapbc/template-infra/pull/282#discussion_r1219930653) that there was a lot of extra code in the configuration scripts. Moreover, @rocketnova has suggested that they'd like to minimize dependencies between infra layers to avoid the brittleness of having to copy the results/outputs of one infrastructure layer into the inputs of another infrastructure layer.

In https://github.com/navapbc/template-infra/pull/159 we came up with an approach to DRY things up by establishing a configuration value like a resource name in a shared configuration module, and then having the downstream infrastructure module use a data source to fetch details about the upstream resource.

However, when I implemented https://github.com/navapbc/template-infra/pull/282 I forgot about that approach.

This PR refactors the code from https://github.com/navapbc/template-infra/pull/282 by getting rid of the database variables from the infra/app/service module, and instead adding values like db_cluster_name and db_access_policy_name to the app-config module and then getting all the values the service needs using terraform data sources.

This doesn't completely get rid of the need for the configure scripts, but it does move us in that direction.

## Testing

* I re-ran `make infra-configure-app-service APP_NAME=app ENVIRONMENT=dev`
  image

* then when i ran `make infra-update-app-service APP_NAME=app ENVIRONMENT=dev` i showed that there are no actual changes to the infrastructure
  image


lorenyu:
Done as part of https://github.com/navapbc/template-infra/pull/321

(Will need to manually update app-config/main.tf since that's on the ignore list in update-template.sh)

PR #18, opened 2023-06-16
Remove unnecessary env vars

lorenyu:
## Ticket

n/a

## Changes
see title

## Context for reviewers
Most of the env vars in run-database-migrations.sh are already defined in the container task definition. The only one that needs to be overwritten is DB_USER, which should be set the user to the migrator user rather than the app user.

## Testing
Ran migrations script locally and it still works. Here are the logs from the migration run:
image


lorenyu:
Done as part of https://github.com/navapbc/template-infra/pull/323

PR #21, opened 2023-06-21
Specify search_path during example app migrations

lorenyu:
## Ticket

n/a

## Changes
see title

## Context for reviewers
PR #327 fixed an issue that wasn't detected on the platform-test repo since platform-test didn't specify the search path when logging in to the database, so the migrations table still created successfully on the `public` schema.

This change specifies the search path to restrict table creations to the app schema as specified by the DB_SCHEMA environment variable

## Testing
Successfully ran deploy job manually from this branch: https://github.com/navapbc/platform-test/actions/runs/5339468699

Ran `make infra-update-app-database APP_NAME=app ENVIRONMENT=dev` to update role manager env vars and then re-ran role manager function with `make infra-update-app-database-roles APP_NAME=app ENVIRONMENT=dev` to make sure it still works:

### DIFF
--- a/app/db-migrate
+++ b/app/db-migrate
@@ -7,7 +7,7 @@ psql \
   --port=$DB_PORT \
   --username=$DB_USER \
   --dbname=$DB_NAME \
-  --set=search_path=$DB_SCHEMA \  
-  --variable=ON_ERROR_STOP=1 \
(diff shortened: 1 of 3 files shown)


lorenyu:
Done as part of https://github.com/navapbc/template-infra/pull/333

PR #25, opened 2023-08-08
Separate DB Access Policies

charles-nava:
## Ticket

[345 Role Separation](https://github.com/navapbc/template-infra/issues/354)

## Changes

- Separated access policy into app_access and migrate_access
- Made new policies available to the database
 
## Testing

- [x] Validate deployment
- [x] Confirm access to app and migrate from database
- [x] Run migrations through make
- [x] Hit migrations endpoint
- [ ] [Improved testing](https://docs.google.com/document/d/1GQQL-xbHXEIfTuLhjfKc0MQKwHUDN-WiAZYTcEvYGF8/edit)


infra/app/app-config/main.tf
```diff
@@ -3,7 +3,7 @@ locals {
   environments                    = ["dev", "staging", "prod"]
   project_name                    = module.project_config.project_name
   image_repository_name           = "${local.project_name}-${local.app_name}"
-  has_database                    = true
+  has_database                    = false
```
lorenyu, line 6: i think we need to keep this as true

infra/app/service/main.tf
```diff
@@ -66,11 +69,6 @@ data "aws_rds_cluster" "db_cluster" {
   cluster_identifier = local.database_config.cluster_name
 }
 
-data "aws_iam_policy" "db_access_policy" {
-  count = module.app_config.has_database ? 1 : 0
-  name  = local.database_config.access_policy_name
-}
-
```
lorenyu, line 73: we need to rename this data source to app_db_access_policy and add a new one migrator_db_access_policy, then pass those in below to the service module

infra/app/service/main.tf
```diff
@@ -88,7 +86,7 @@ module "service" {
 
   db_vars = module.app_config.has_database ? {
     security_group_ids = data.aws_rds_cluster.db_cluster[0].vpc_security_group_ids
-    access_policy_arn  = data.aws_iam_policy.db_access_policy[0].arn
```
lorenyu, line 91: we need to pass in the app db access policy and migrator db access policy to the service module

lorenyu (commented):
one more thing i forgot to mention. in order for the db-migrate task to be able to log into db as the migrator user, the db-migrate task needs the migrator role. by default, the task definition assigns the [service role to the task](https://github.com/navapbc/platform-test/blob/main/infra/modules/service/main.tf#L161), and the [service role has attached to it the app_db_access policy](https://github.com/navapbc/platform-test/blob/main/infra/modules/service/main.tf#L394-L399). so in [run-database-migrations.sh](https://github.com/navapbc/platform-test/blob/main/bin/run-database-migrations.sh), we're going to need to override the executionRoleArn to be the migrator role, not the service role (see [overrides in run-task aws cli docs](https://docs.aws.amazon.com/cli/latest/reference/ecs/run-task.html)), and we're going to have to attach the migrator_db_access_policy to the migrator role. Note that in order to do this, we'll also need to update [run-command.sh](https://github.com/navapbc/platform-test/blob/main/bin/run-command.sh) to accept role overrides (currently it only accepts environment variables as overrides).

charles-nava:
As discussed Infra Service checks can be ignored due to changes affecting multiple layers causing failure

charles-nava:
These changes will be brought over in a multistep deployment starting with the branch `db-role-separate-A-additive`

PR #26, opened 2023-08-10
Explain how to fix default table permissions in app

lorenyu:
## Ticket

## Context for reviewers

see context in https://github.com/navapbc/template-infra/pull/377

## Testing

Before the /migrations endpoint was getting a 500 internal server error and the logs showed:
image
image

Afterwards:
image


lorenyu:
Done as part of https://github.com/navapbc/template-infra/pull/377/files

PR #30, opened 2023-08-23
Refactor infra tests to prepare for db tests

lorenyu:
## Ticket

## Changes

* Add helpers.go file for helper functions for tests
* Create TerraformInit helper function
* Rename TestDev to TestService
* Separate service e2e infra tests into its own ci-infra-service.yml workflow
* Rename make infra-test to make infra-test-service

## Context for reviewers

This is in preparation to add db tests. Databases take too long to spin up and down so I want to have separate CI job for db that only runs when there are changes to db files.

## Testing

CI should still pass after refactor


lorenyu:
Done with https://github.com/navapbc/template-infra/pull/388

PR #36, opened 2023-08-25
Migrate VPC endpoints to separate networks module (step 2/2)

lorenyu:
## Ticket

This is work done in prep for https://github.com/navapbc/template-infra/issues/277

## Changes

* Replace VPC endpoints in database module with data sources referencing VPC endpoints created in network module

## Context for reviewers

This is the second part of migrating VPC endpoints to the network module from the database module

## Rollout

As part of rolling out https://github.com/navapbc/template-infra/pull/405 we imported VPC endpoints to a separate networks module, so as part of rolling out this module, we need to [terraform rm](https://developer.hashicorp.com/terraform/cli/commands/state/rm) the VPC endpoints from the database module before applying on platform-test and platform-test-flask:

```bash
terraform -chdir=infra/app/database state rm module.database.aws_vpc_endpoint.kms
terraform -chdir=infra/app/database state rm module.database.aws_vpc_endpoint.ssm
make infra-update-app-database APP_NAME=app ENVIRONMENT=dev
```

## Testing

- [x] Create a new terraform workspace for database layer
   
   image

- [x] Create database cluster in new workspace

   image

- [x] Run database role manager to make sure it still works (the role manager is what needed the VPC endpoints)

   image

   (the roles prefixed with lyvpce were created before I fixed the app and migrator names not to be prefixed with the workspace name)

- [x] Clean up by destroying the database and workspace

   first disable deletion protection
   image

   then run `terraform -chdir infra/app/database destroy -var-file=dev.tfvars`
   image

   now clean up the workspace
   image

### DIFF
--- a/.github/workflows/cd-app.yml
+++ b/.github/workflows/cd-app.yml
@@ -1,7 +1,7 @@
-name: Continuous Deployment - App
+name: Deploy App
 # Need to set a default value for when the workflow is triggered from a git push
 # which bypasses the default configuration for inputs
-run-name: Deploy ${{ github.ref_name }} to ${{ inputs.environment || 'dev' }}
+run-name: Deploy ${{ github.ref_name }} to App ${{ inputs.environment || 'dev' }}
 
 on:
   push:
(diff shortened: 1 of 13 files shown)


lorenyu:
Done as part of https://github.com/navapbc/template-infra/pull/406

PR #38, opened 2023-08-28
App to use specific role for app database actions

charles-nava:
## Ticket

Part of [354 Separation of Database Roles](https://github.com/navapbc/template-infra/issues/354)

## Changes

- Adds app and migration specific roles
- Changes app settings to use app specific IAM role

## Context for reviewers

Part 2 of [multipart update ](https://github.com/navapbc/template-infra/issues/354#issuecomment-1693973424)

## Testing

- [x] Confirm successfully deploys
- [x] Confirm correct role is attached to the ECS task


infra/modules/service/database-access.tf
```diff
@@ -14,9 +14,24 @@ resource "aws_vpc_security_group_ingress_rule" "db_ingress_from_service" {
   referenced_security_group_id = aws_security_group.app.id
 }
 
+
```
lorenyu, line 17: optional nit: looks like an extraneous newline

infra/modules/service/database-access.tf
```diff
@@ -14,9 +14,24 @@ resource "aws_vpc_security_group_ingress_rule" "db_ingress_from_service" {
...
+  count = var.db_vars != null ? 1 : 0
+
+  role       = aws_iam_role.app_service.name
+  policy_arn = var.db_vars.app_access_policy_arn
+}
+
+resource "aws_iam_role_policy_attachment" "migrator_db_access" {
+  count = var.db_vars != null ? 1 : 0
+
+  role       = aws_iam_role.migrator_service.name
```
lorenyu, line 35: I didn't catch this in the other PR, but we shouldn't name the role `migrator_service`. the migrations are run as a one-off task, not as a service. can you rename the role to something like `migrator` or `migrator_task`?

infra/modules/service/database-access.tf
```diff
@@ -14,9 +14,24 @@ resource "aws_vpc_security_group_ingress_rule" "db_ingress_from_service" {
   referenced_security_group_id = aws_security_group.app.id
 }
 
+
 resource "aws_iam_role_policy_attachment" "app_db_access" {
   count = var.db_vars != null ? 1 : 0
 
   role       = aws_iam_role.service.name
   policy_arn = var.db_vars.access_policy_arn
 }
```
lorenyu, line 23: when are we planning on removing this block?

optional idea: what if we attached the migrator access policy to the app_service role rather than to the migrator role in the `migrator_db_access` resource below? then we can remove this policy attachment.

in the next PR when we update run-command.sh and run-database-migrations.sh to override the migration task to use the migrator role, we can then update the migrator_db_access policy attachment to attach to the migrator role instead of the app role.

in either case, an optional idea to add a # TODO comment of what we need to do. if we stick with the current PR's approach add a "TODO: delete this app_db_access policy attachment once we've successfully updated migrations to use the migrator role", or if we go with the idea in this comment, we'd add TODO in the migrator_db_access resource that says "TODO: update role to be migrator role once we update run-database-migrations.sh to override the task role to be the migrator role"

lorenyu (commented):
Code looks good but I don't see evidence of testing. Are you going to spin up a cluster in infra/app/service in a new workspace? Then you can show that it's healthy and hit the /migrations endpoint – that would be one way of testing this.

infra/modules/service/database-access.tf
charles-nava: Removed

infra/modules/service/database-access.tf
charles-nava: Sounds good and done

infra/modules/service/database-access.tf
charles-nava: I was planning on removing it as part of step 3. So next PR. For this PR, the original role is still used for migration tasks
https://github.com/navapbc/template-infra/issues/354#issuecomment-1693973424

lorenyu: i think you'll find that that won't work. because in this PR the task definition is set to the app_service role, but the app_service role only has the app_access_policy. that means when we call run-database-migrations.sh, the migration script will try to log in as the migrator user but since the task role only has the app_service policy it won't be able to.

charles-nava: Added todo comments

charles-nava:
Screenshot 2023-08-29 at 12 27 42 PM
Screenshot 2023-08-29 at 12 30 58 PM
Screenshots of successful deployment and task definition

charles-nava:
> Code looks good but I don't see evidence of testing. Are you going to spin up a cluster in infra/app/service in a new workspace? Then you can show that it's healthy and hit the /migrations endpoint – that would be one way of testing this.

Testing results in comment now

charles-nava:
Screenshot 2023-08-29 at 3 45 29 PM
`make release-run-database-migrations` was successful, with correct timestamp in endpoint
Screenshot 2023-08-29 at 3 49 26 PM

infra/modules/database/authentication.tf
```diff
@@ -38,6 +41,8 @@ data "aws_iam_policy_document" "app_db_access" {
 
     resources = [
       "${local.db_user_arn_prefix}/${var.app_username}",
+      # TODO: Remove migrator username. Part 3 of multipart update https://github.com/navapbc/template-infra/issues/354#issuecomment-1693973424
+      "${local.db_user_arn_prefix}/${var.migrator_username}",
```
lorenyu, line 45: doing it this way will break things and defeats the purpose of us doing the db layer changes first, since this requires you to apply in db layer first before applying in service layer, but we have CD that autodeploys the service.

i think you need to do [the idea i suggested in this PR comment](https://github.com/navapbc/platform-test/pull/38#discussion_r1308076465):

> what if we attached the migrator access policy to the app_service role rather than to the migrator role in the migrator_db_access resource below? then we can remove this policy attachment.

in other words, don't change the policies that were created in the db layer. just make sure you attach both the app access policy and the migrator access policy to the service.

lorenyu (commented):
this approach doesn't work since it has changes to the db layer. remember, the approach is:

1. backwards compatible db layer changes (new policies)
2. use new policies in service layer
3. use new policies for migrator task in service layer
4. remove old policies

infra/modules/service/access-logs.tf
```diff
@@ -21,7 +21,7 @@ locals {
 
 resource "aws_s3_bucket" "access_logs" {
   bucket_prefix = "${var.service_name}-access-logs"
-  force_destroy = false
+  force_destroy = true
```
lorenyu, line 24: no biggie since this is just on the platform-test PR that won't get merged but looks like you accidentally checked this in

infra/modules/service/access-control.tf
lorenyu, line 1: aws_iam_role.service isn't used anymore and can be deleted

lorenyu (commented):
looks good.
- still needs some cleaning up, there are two resources that aren't used anymore that can be deleted: aws_iam_role.service and aws_iam_role_policy_attachment.app_db_access
- i still see a change in the database layer
- there's a leftover force_destroy change on the access logs bucket probably from local testing

infra/modules/service/access-logs.tf
charles-nava: Reverted. Good catch

infra/modules/service/database-access.tf
```diff
@@ -14,9 +14,24 @@ resource "aws_vpc_security_group_ingress_rule" "db_ingress_from_service" {
...
+
+resource "aws_iam_role_policy_attachment" "app_service_db_access" {
+  count = var.db_vars != null ? 1 : 0
+
+  role       = aws_iam_role.app_service.name
+  policy_arn = var.db_vars.app_access_policy_arn
+}
```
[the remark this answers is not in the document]
charles-nava: Madee the change

infra/app/database/main.tf
lorenyu, line 1: @charles-nava did you accidentally revert this file to an old version?

lorenyu (commented):
looks good now but i see a random change to infra/app/database/main.tf that looks like a reverting to an old version before the vpc endpoint change i made. curious how that happened. let's fix that and then do a final round of testing.

infra/app/database/main.tf
charles-nava: Not sure what happened with this one but it should be taken care of now

charles-nava: Marked this as unresolved due to the fix not showing up in the PR despite being [evident in the branch](https://github.com/navapbc/platform-test/blob/1c30a23fe3319d2ce4052aa717f2343446ed57ff/infra/app/database/main.tf#L61)

charles-nava: Yeah I think github just lagged out. It looks like it's good to go on this front

charles-nava:
> looks good now but i see a random change to infra/app/database/main.tf that looks like a reverting to an old version before the vpc endpoint change i made. curious how that happened. let's fix that and then do a final round of testing.

I think it was likely due to copying changes over from the `template-infra` before merging main into there

lorenyu (commented):
Looks good. Make sure to retest (the steps you did previously to test are still appropriate) and if everything passes we can move the changes over to template-infra.

re: accidentally copying over old changes from main, going forward make sure to review your commits and also review your own PRs before requesting others to review.

charles-nava:
Screenshot 2023-09-06 at 3 25 13 PM
Screenshot 2023-09-06 at 3 23 59 PM
Screenshot 2023-09-06 at 3 21 47 PM
Testinc complete

charles-nava:
Handled in alternate PR

PR #40, opened 2023-09-01
Add ability to check database roles configuration

lorenyu:
## Ticket

n/a

## Changes
* Add "check" option to role manager lambda function
* Add check-database-roles.sh script to call role manager with "check" option
* Add make infra-check-database-roles target
* Add cluster_id output to database module
* Add db app and migrator access to role manager lambda function
* Update set-up-database.md instructions to check database roles at the end of setup

Other changes
* Print role manager's logs when running create-or-update-database-roles.sh

## Context for reviewers

Another developer experience thing similar to https://github.com/navapbc/template-infra/pull/415, this change allows someone to check that the database is properly configured without having to spin up the service layer. This change updates the role manager lambda function to have a "check" mode that logs in to the database as the "migrator" and "app" users (to make sure that IAM auth works), and attempts to perform queries that on the database to check that they have proper privileges. Specifically, the migrator user checks the search path to make sure that new tables are created in the right schema rather than public, it creates a table, then the app user writes something to that table, then the migrator user drops the table. This should improve the first time developer setup experience, and should also help with maintenance e.g. if someone needs to refactor the database code, they can do so and have some confidence that their changes won't break things.

## Testing

1. Create new database cluster in a new workspace `lycheckdb`
   image

   image

2. Run database check function (this should fail since we didn't run configure database roles function yet)

   image

3. Run configure database roles function
   image

4. Run database check function again

   first time timed out
   image

   retry was successful
   image

5. Clean up (destroy db and delete workspace)

TODO once we're done with this PR


lorenyu:
Resolved conflicts, switched to new using new split policies (rather than the old policy that will be deleted), and retested:

image

image

(note: checking roles took two tries, first try timed out)

PR #41, opened 2023-09-05
Consolidate build-repository config

lorenyu:
## Ticket

Work for https://github.com/navapbc/template-infra/issues/301

## Changes

* Add ADR for consolidating infra config into app-config module
* Add account-ids-by-name.sh utility script for mapping account names to account ids
* Add build-repository config in app-config module
* Remove configure-app-build-repository script and update make infra-configure-app-build-repository command
* Remove build-repository's example.tfvars and terraform.tfvars files

## Context for reviewers

Revisiting earlier feedback from @kyeah that config is spread out across multiple files is error prone and from @acouch that there's a lot of duplicate code in the configure scripts, this PR proposes a method for consolidating the configs all into the app-config module.

This PR just changes the build-repository module. PR #42 consolidates the infra configuration for the app/database and app/service modules.

## Testing

Ran `make infra-configure-app-build-repository APP_NAME=app` which should be a no-op since the command should be idempotent and the build repository is already configured.

image

Then I ran `make infra-update-app-build-repository APP_NAME=app` which should show a no-op plan

image


bin/account-ids-by-name.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
rocketnova, line 1: Gathering the account names and account IDs from filenames _feels_ brittle to me, but the only better idea I have is to save these to a separate config file and I'm sure if that's better? 🤷 I can just imagine that someone changes the filename, not realizing that the structure of the filename is encoding data that gets used elsewhere.

infra/app/build-repository/main.tf
```diff
@@ -9,6 +9,12 @@ locals {
     application_role = "build-repository"
     description      = "Backend resources required for storing built release candidate artifacts to be used for deploying to environments."
   })
+
+  # Get list of AWS account ids for the application environments that
+  # will need access to the build repository
+  app_account_names   = values(module.app_config.account_names_by_environment)
+  account_ids_by_name = data.external.account_ids_by_name.result
+  app_account_ids     = [for account_name in local.app_account_names : local.account_ids_by_name[account_name] if contains(keys(local.account_ids_by_name), account_name)]
```
rocketnova, line 17: I don't totally follow this part, but I haven't used AWS in a multi-account scenario, so I might not be understanding why we need to gather account IDs in this way.

docs/decisions/infra/0008-consolidate-infra-config-from-tfvars-files-into-config-module.md
```diff
@@ -0,0 +1,29 @@
...
+## Context
+
+Currently, application infrastructure configuration is split across config modules (see [app-config](/infra/app/app-config/)) as well as .tfvars files in each of the application's infra layers - infra/app/build-repository, infra/app/database, and infra/app/service. As @kyeah pointed out, it’s easy to make mistakes when configuration is spread across multiple files, and expressed a desire to manage manage tfvars across environments all in a single file the way that some applications do for application configuration. And as @acouch [pointed out](https://github.com/navapbc/template-infra/pull/282#discussion_r1219930653), there is a lot of duplicate code with the configure scripts (setup-current-account.sh, configure-app-build-repository.sh, configure-app-database.sh, configure-app-service.sh) that configure the backend config and variable files for each infrastructure layer, which increases the burden of maintaining the configuration scripts.
+
+## Overview
+
+This ADR proposes the following:
+
+* Move all environment configuration into [app-config](/infra/app/app-config/) modules
+* Remove the need for .tfvars files
```
rocketnova, line 18: Is the idea to remove all `.tfvars` files? I think they were useful to have to be able to have different values for different environments.

rocketnova (commented):
2 non-blocking questions, but the question about `.tfvars` files should be clarified before moving forward.

docs/decisions/infra/0008-consolidate-infra-config-from-tfvars-files-into-config-module.md
rocketnova: Clarification: I think it's useful to have a `.tfvars` for the app-config, but I think removing the need for them in the other modules is a good improvement.

infra/app/build-repository/main.tf
lorenyu: yeah. if a project has a separate AWS account for each environment, then we need to set the IAM policy on the ECR to allow cross-account access, so we need a list of the accounts ids for this application. we have a map from environments to account names in the app-config, so we need to map the account names to account ids to get the list of account ids for all the environments for this application.

bin/account-ids-by-name.sh
lorenyu: yeah i had the same thought.

digging deeper into the past here, i think we wanted both:
- a way to go from the current account id that the developer is currently authenticated with (which we can get from aws sts get-caller-identity) to the tfbackend config file so that we can do `terraform init -backend-config=<current account's config>`
- we wanted the tfbackend config filename to be human readable e.g. dev.s3.tfbackend rather than <account id>.s3.tfbackend

so we decided to name the files `<human readable account name>.<account id>.s3.tfbackend`, then in our makefile we get the current backend config filename by searching for the current account id amongst the filenames (see [make infra-update-current-account](https://github.com/navapbc/template-infra/blob/90b7e18433eb21a3f46f24890c473b0849e27d49/Makefile#L90-L91) and [current-account-config-name.sh](https://github.com/navapbc/template-infra/blob/90b7e18433eb21a3f46f24890c473b0849e27d49/bin/current-account-config-name.sh))

given that we already have both the account name and account id in the filenames, i thought having a separate config file mapping account names to account ids would be less DRY.

an alternative would be to name the files `dev.s3.tfbackend`, without the account id, and have the account id in a separate file like `account_ids_by_name.json`, and then whenever we run `set-up-current-account` with a new account, the script would update the `account_ids_by_name.json` file. and to get the account name from the account id, we'd invert the map using jq like `cat account_ids_by_name.json | | jq 'to_entries | .[] | {(.value): .key}` and then use the resulting inverse map to get the account name by account id.

if the alternative is not much better, for now, i could also just add comments to all the places that rely on this filename structure:

- current-account-config-name.sh
- account-ids-by-name.sh
- .github/acounts/configure-aws-credentials/action.yml

thoughts?

docs/decisions/infra/0008-consolidate-infra-config-from-tfvars-files-into-config-module.md
lorenyu: app-config isn't a root module, so i'm not sure how we'd have `.tfvars` files for that module. could you elaborate on what you had in mind?

bin/account-ids-by-name.sh
rocketnova: 🤔 That's a tough one. I'm leaning towards the alternative that you described. I think that keeping the account IDs in a separate file is easier to approach.

infra/app/build-repository/main.tf
rocketnova: Oh interesting. I had forgotten about the situation where you might want the ECR to be in one account, but then deploy from it to a complete environment (without ECR) in another. Got it.

bin/account-ids-by-name.sh
lorenyu: Mind if I push that off to a separate PR? It'll be a breaking change since I have to first copy over the <name>.<id>.s3.tfbackend files to <name>.tfbackend files and generate the separate config so that the scripts won't break when I update them, then update the scripts and finally remove the old files.

I created a ticket for it: https://github.com/navapbc/template-infra/issues/424

docs/decisions/infra/0008-consolidate-infra-config-from-tfvars-files-into-config-module.md
rocketnova: Sorry about that. I think I'm following what's happening in these PRs a little bit better. Part of my confusion stemmed from me trying to follow the differences between the BDS implementation that I think Sawyer did and what's in #42. 

Is the idea that all the variables for `build-repository`, `database`, and `service` should all be the same across environments? 

If a project wanted to make them different in one environment, how would they accomplish this in the proposed model?

bin/account-ids-by-name.sh
rocketnova: Of course!

docs/decisions/infra/0008-consolidate-infra-config-from-tfvars-files-into-config-module.md
lorenyu: > Is the idea that all the variables for build-repository, database, and service should all be the same across environments?
> 
> If a project wanted to make them different in one environment, how would they accomplish this in the proposed model?

the build-repository layer only has one environment called "shared", since the build-repository is shared across all the service environments

for the database and service layer, which might have dev, staging, prod environments, it's not in the template now but in a separate PR I was going to update the template's app-config to demonstrate how one would change things per environment.

currently the app-config module is set up like this:

```
locals {
  environments                    = ["dev", "staging", "prod"]
  environment_configs = { for environment in local.environments : environment => module.env_config[environment] }
}
module "env_config" {
  for_each = toset(local.environments)

  source                          = "./env-config"
  app_name                        = local.app_name
  environment                     = each.key
  has_database                    = local.has_database
  has_incident_management_service = local.has_incident_management_service
}
```

the anticipated way that most projects will actually do this is to have the environment configs listed out separately e.g.

```
locals {
  environments                    = ["dev", "staging", "prod"]
  environment_configs = {
    dev = module.env_config_dev
    staging = module.env_config_staging
    prod = module.env_config_prod
  }
}
module "env_config_dev" {
  source                          = "./env-config"
  app_name                        = local.app_name
  environment                     = "dev"
  has_database                    = local.has_database
  has_incident_management_service = local.has_incident_management_service
}
module "env_config_staging" {
  source                          = "./env-config"
  app_name                        = local.app_name
  environment                     = "staging"
  has_database                    = local.has_database
  has_incident_management_service = local.has_incident_management_service
}
module "env_config_prod" {
  source                          = "./env-config"
  app_name                        = local.app_name
  environment                     = "prod"
  has_database                    = local.has_database
  has_incident_management_service = local.has_incident_management_service
}
```

so if a project would want to add a new config variable (e.g. CPU count), they'd add a variable (cpu_count) to the env-config module, optionally provide a default value, and then in app-config/main.tf you can specify which environment has which value.

rocketnova: Have you considered controlling environment config via different `.tfvars` files (e.g. `dev.tfvars`, `staging.tfvars`) in either app_config or env_config? That way if a project ends up wanting to expose a lot of different config on different environments as the project grows, it's easy to `diff` files than to compare across lines in the same file

lorenyu: The point of this PR was actually to move away from having configs spread out across different layers and files (e.g. infra/app/service/(dev|staging|prod).tfvars, infra/app/database/(dev|staging|prod).tfvars) and into one place (infra/app/app-config).

If the goal is to make it easy to diff, you can do that without using .tfvars by just splitting things up into separate .tf files. My previous example would become

```
# dev.tf
module "env_config_dev" {
  source                          = "./env-config"
  app_name                        = local.app_name
  environment                     = "dev"
  has_database                    = local.has_database
  has_incident_management_service = local.has_incident_management_service
}
```
```
# staging.tf
module "env_config_staging" {
  source                          = "./env-config"
  app_name                        = local.app_name
  environment                     = "staging"
  has_database                    = local.has_database
  has_incident_management_service = local.has_incident_management_service
}
```
```
# prod.tf
module "env_config_prod" {
  source                          = "./env-config"
  app_name                        = local.app_name
  environment                     = "prod"
  has_database                    = local.has_database
  has_incident_management_service = local.has_incident_management_service
}
```

it's actually not possible to put separate .tfvars in the app-config module since it's not a root module. it's a shared config child module that's used by all the other root modules (infra/app/build-repository, infra/app/service, infra/app/database).

rocketnova: Oh right. I forget that you want to have separate terraform states per module. I definitely don't want to drag out this process more, so don't wait on me to approve to merge.

Can you remind me why we chose to have separate terraform modules per module (e.g. database, service) and not one for the entire app? Is it generally better practice to run terraform on sub-portions of an environment rather than the entire environment altogether? I have tended to optimize for being able to run `terraform apply` once to be able to get an entire environment into the correct state.

lorenyu: I could have sworn we had an ADR for this, but I can't find it.

I used to have the same preference as what you described (everything in one tfstate file) but after discussions with Brendan Neutra and how they evolved their infra architecture thinking over time on the federal projects he worked on, I saw the value of splitting things up by layer. Some of the reasons include:

* Less risky – some layers have changes applied on every merge to main e.g. the service layer needs to update the image_tag in the task definition on every deploy. On layers you don't expect to have frequent changes and where changes have higher risks, such as your database layer, it's cleaner to have the terraform apply -auto-approve not be able to apply to those layers
* Differences in how each layer treats "environments" – Not all layers have the same concept of "environment" as the application. The AWS account layer might have one account for everything, one account for nonprod and prod, or one account per environment. The network layer (VPC) can have similar variations (one VPC for non-prod, one for prod, or one per environment). The build repository layer only has one instance that's shared across all environments. Putting all resources in one root module would disallow these variations between layers unless you introduce differences between environments (e.g. you add logic like "if environment = prod create build-repository"), but then by having different environments have diverging configuration logic, that defeats the purpose of having them share the terraform module files.
* AWS uniqueness constraints – kind of a special case of the previous bullet, but some resources have uniqueness constraints. Like you can only have one OIDC provider for github actions per AWS account. You can only have one VPC endpoint per VPC per AWS service.
* Authorization constraints – In some projects, the agency has restrictions on the ability to create resources in certain layers (e.g. VPCs). So separating them allows us to continue to manage downstream layers even if upstream layers have already been created externally.
* Performance – when you have a lot of resources in your tfstate file, terraform needs to make many calls to AWS just to check the current state of the infrastructure. Making a ton of AWS calls for things that rarely ever change causes terraform applies to get slower and slower

rocketnova: Ooo very helpful. Thank you! I think it would be helpful to capture this as either an ADR or in the `docs/infra` (my preference would probably be both tbh). It helps explain the reasoning behind this structure.

rocketnova (approved):
After much discussion, I'm on board with this decision. I think it would be helpful to capture some of this in `docs/infra`.

docs/decisions/infra/0008-consolidate-infra-config-from-tfvars-files-into-config-module.md
lorenyu: Yeah I agree, I created a ticket for it and will work on that as soon as I can https://github.com/navapbc/template-infra/issues/431

lorenyu:
Done in https://github.com/navapbc/template-infra/pull/432

PR #42, opened 2023-09-05
Consolidate database and service configs

lorenyu:
## Ticket

Work for https://github.com/navapbc/template-infra/issues/301

## Changes

* Move database's region to database_config in app-config module
* Fetch service's tfstate_bucket and tfstate_key values directly from backend config file
* Add service_config in app-config module
* Remove configure-app-database script and update make infra-configure-app-database command
* Remove configure-app-service script update make infra-configure-app-service command
* Remove app/database and app/service layer's example.tfvars and dev.tfvars files

## Context for reviewers

This PR builds on #41 to consolidate infra config into the app-config module, and remove the configure scripts and .tfvars files

## Testing

Ran make infra-configure-app-database APP_NAME=app ENVIRONMENT=dev which should be a no-op since the command should be idempotent and the build repository is already configured.

image

Then I ran make infra-update-app-database APP_NAME=app ENVIRONMENT=dev which should show a no-op plan

image

Ran make infra-configure-app-service APP_NAME=app ENVIRONMENT=dev which should be a no-op since the command should be idempotent and the build repository is already configured.

image

Then I ran make infra-update-app-service APP_NAME=app ENVIRONMENT=dev which should show a no-op plan

image


Makefile
rocketnova, line 1: I really like these changes 👍

infra/app/service/image_tag.tf
```diff
@@ -6,6 +6,15 @@
 #  1. Accept an optional variable during a terraform plan/apply. (see "image_tag" variable in variables.tf)
 
 #  2. Read the output used from the last terraform state using "terraform_remote_state".
+#     Get the backend config by parsing the backend config file
+locals {
+  backend_config_file_path = "${path.module}/${var.environment_name}.s3.tfbackend"
+  backend_config_file      = file("${path.module}/${var.environment_name}.s3.tfbackend")
+  backend_config_regex     = "(\\w+)\\s*=\\s*\\\"(\\w+)\\\""
+  backend_config           = { for match in regexall("(\\w+)\\s*=\\s*\\\"(.+)\\\"", local.backend_config_file) : match[0] => match[1] }
```
rocketnova, line 14: 👀 👀 These regexes. Can they be avoided at all? If no, can they be documented?

rocketnova (commented):
I think the comment in https://github.com/navapbc/platform-test/pull/41#discussion_r1316438074 about if the idea is to remove all `.tfvars` may still be relevant here.

infra/app/service/image_tag.tf
lorenyu: yeah i'll add a comment.
the regex itself isn't that bad, but terraform unfortunately doesn't have native way to write regexes (unlike in javascript or python), so we have to double escape them. the regex is actually just:

```
(\w+)\s*="\s*(.+)"
```

also i realized i typed the regex twice accidentally and i overescaped the quotes

lorenyu: re: whether we can avoid them: 
if you know of a native terraform function that can parse tfvars files that'd be great, but i couldn't find one

lorenyu: Ok simplified the regex and added a comment. Do you think that makes it clearer?

rocketnova: Yup. Looks great! Thanks

rocketnova (approved):
LGTM! We have resolved the previously referenced discussion in #41 .

lorenyu:
Just ran through my test plan and uploaded evidence in the description. I found one update to the Makefile that I missed but fixed that.

lorenyu:
Done in https://github.com/navapbc/template-infra/pull/433

PR #45, opened 2023-09-06
Use correct roles

lorenyu:
## Ticket

## Changes

N/A

## Context for reviewers

We forgot to pass in the correct policies in this PR https://github.com/navapbc/template-infra/pull/411 

## Testing

Since it was late in the day and no one else was using the dev environment I just did my testing in the default workspace.

I first applied the changes to service layer with `terraform -chdir=infra/app/service apply -var-file=dev.tfvars`

image

Then I ran deploy from this branch


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/429

PR #47, opened 2023-09-07
Add option to override task role in run-command.sh

charles-nava:
Part of separation of db roles #354
Uses while loop and case switch to allow for definition of vars on flags

## Ticket

Related to #354

## Changes

* Refactor environment variables parameter for run-command to use --environment-variables optional flag
* Add --task-role-arn option to override task role

## Context for reviewers

> Necessary for separation of database roles

## Testing

- [ ] Successful deploy
- [ ] Run db-migration command with app role 
- [ ] Run db-migration command with migrate role


bin/run-command.sh
```diff
@@ -18,10 +18,101 @@ set -euo pipefail
...
 
-APP_NAME="$1"
-ENVIRONMENT="$2"
-COMMAND="$3"
-ENVIRONMENT_VARIABLES=${4:-""}
+ACCOUNT_ID="$(./bin/current-account-id.sh)"
+
+while :; do
+  case $1 in
+    -a|--app|--app-name)
```
lorenyu, line 25: let's keep it simple and just provide one flag option with it full spelled out

bin/run-command.sh
```diff
@@ -18,10 +18,101 @@ set -euo pipefail
...
+      if [ "$2" ]; then
+        APP_NAME=$2
+        shift
+      else
+        die "$1 flag present without app name"
+      fi
+      ;;
+    -e|--env|--env-name|--environment)
```
lorenyu, line 33: app name, environment, and command aren't optional, could we keep them as positional parameters?

lorenyu (commented):
didn't do a full review yet but hopping into some meetings so leaving a quick comment to unblock you

bin/run-command.sh
charles-nava: It would add more complication to handle both positional and flag based arguments but I can give it a shot

bin/run-command.sh
charles-nava: Can do but this is still in progress. Don't feel pressured to monitor and fix this

bin/run-command.sh
lorenyu: If it's more complicated then we can keep it the way it is. I didn't know if there was a simple way to handle both

lorenyu: what about something like

```
while :; do
  case $1 in
    --role-name)
      ...
    --environment-variables)
      ...
    *)
      APP_NAME=$1
      ENVIRONMENT=$2
      COMMAND=$3
      break
      ;;
done
```

lorenyu: Just tested it out, this seems to work:

```bash
while :; do
  case $1 in
    --environment-variables)
      ENVIRONMENT_VARIABLES=$2
      shift 2
      ;;
    --role-name)
      ROLE_NAME="$2"
      shift 2
      ;;
    *)
      APP_NAME="$1"
      ENVIRONMENT="$2"
      COMMAND="$3"
      break
      ;;
  esac
done

echo "APP_NAME=$APP_NAME"
echo "ENVIRONMENT=$ENVIRONMENT"
echo "COMMAND=$COMMAND"
echo "ENVIRONMENT_VARIABLES=${ENVIRONMENT_VARIABLES:-}"
echo "ROLE_NAME=${ROLE_NAME:-}"
```

image

charles-nava: I'll give that a shot after I figure out some of the other issues

bin/run-command.sh
```diff
@@ -18,10 +18,71 @@ set -euo pipefail
...
+          shift 2
+        else
+          echo "$1 flag present without role name"
+          exit 1
+        fi
+      else
+        printf "Conflicting attempts to define role\n%s\n%s" "$ROLE_ARN" "$1"
+        exit 1
+      fi
+      ;;
```
lorenyu, line 51: to begin with we'll only have one place that's using the role override, so we only need one option. having two options means we'll have unreachable/dead code, and unreachable code is bad to keep around since it creates a maintenance burden without adding any value. a useful way to think in general is that you should treat all code as a liability and minimize the amount of code we need to accomplish what we're trying to do.

bin/run-command.sh
```diff
@@ -18,10 +18,71 @@ set -euo pipefail
...
+    --role-arn)
+      if [ -z ${ROLE_ARN+x} ]; then
```
lorenyu, line 53: this check is unnecessary. i don't see value in checking for the option being present multiple times. in fact even heavily used shell utilities have the behavior of treating the last option as the one that is respected. this allows people to compose things better.

bin/run-command.sh
```diff
@@ -18,10 +18,71 @@ set -euo pipefail
...
+        if [ "$2" ]; then
```
lorenyu, line 54: this check, while well intentioned, doesn't actually check what it should be checking. if we add the option but forget to add a value and just move on to the next option, it doesn't catch the right error. see screenshot below for an example.

image

you may want to look into `get-opts` (see https://www.baeldung.com/linux/bash-parse-command-line-arguments)

lorenyu (commented):
left some more feedback on the error handling. you may want to look into getopts

charles-nava:
[Logs of run using app role (temporarily given migrator access)](https://github.com/navapbc/platform-test/files/12561397/app_role_test.txt)
[Logs of run using migrator role](https://github.com/navapbc/platform-test/files/12561400/migrator_role_test.txt)
[Logs of run with no role specification](https://github.com/navapbc/platform-test/files/12561401/no_role_test.txt)

bin/run-command.sh
charles-nava: Sounds good. Do we think that role-arn or role-name makes more sense in this instance? Either should be pretty easy to get programmatically, role-name is slightly more human friendly. ARN's are a bit more "standard"

bin/run-command.sh
charles-nava: Sounds good. I'll drop those checks

bin/run-command.sh
charles-nava: Good catch

bin/run-command.sh
lorenyu: let's go with role-arn

bin/run-command.sh
charles-nava: Resolved in e8bd95c

bin/run-command.sh
lorenyu: I looked into getopts, but it only handles single character arguments, and the GNU command get-opt seems a bit more complicated than it's worth. I would be open to just not do error handling for this

charles-nava:
Updated test logs:
[test_app_role.txt](https://github.com/navapbc/platform-test/files/12563930/test_app_role.txt)
[test_migrator_role.txt](https://github.com/navapbc/platform-test/files/12563931/test_migrator_role.txt)
[test_no_role.txt](https://github.com/navapbc/platform-test/files/12563932/test_no_role.txt)
[test_validation_check.txt](https://github.com/navapbc/platform-test/files/12563933/test_validation_check.txt)

lorenyu:
Testing: I copied the test file that Charles gave me:

**Test file**

```bash
APP_NAME=app
ENVIRONMENT=dev
COMMAND='["db-migrate"]'
ENVIRONMENT_VARIABLES='[{ "name" : "DB_USER", "value" : "migrator" }]'
TASK_ROLE_ARN='arn:aws:iam::430004246987:role/app-dev-migrator'

./bin/run-command.sh --task-role-arn "arn:aws:iam::430004246987:role/app-dev-app" --environment-variables "$ENVIRONMENT_VARIABLES" "$APP_NAME" "$ENVIRONMENT" "$COMMAND" > app_role_test.txt
./bin/run-command.sh --task-role-arn "arn:aws:iam::430004246987:role/app-dev-migrator" --environment-variables "$ENVIRONMENT_VARIABLES" "$APP_NAME" "$ENVIRONMENT" "$COMMAND" > migrator_role_test.txt
./bin/run-command.sh --environment-variables "$ENVIRONMENT_VARIABLES" "$APP_NAME" "$ENVIRONMENT" "$COMMAND" > no_role_test.txt
```

And here are the results of that test:

[app_role_test.txt](https://github.com/navapbc/platform-test/files/12580523/app_role_test.txt)
[migrator_role_test.txt](https://github.com/navapbc/platform-test/files/12580524/migrator_role_test.txt)
[no_role_test.txt](https://github.com/navapbc/platform-test/files/12580525/no_role_test.txt)

lorenyu:
Done in https://github.com/navapbc/template-infra/pull/440

PR #50, opened 2023-09-08
Fix deploy release

lorenyu:
## Ticket

n/a

## Changes

see title

## Context for reviewers

This fix https://github.com/navapbc/template-infra/pull/435/files fixed migrations, but now the last step of the deploy (deploy release) fails for the same reason that the migrations failed (see failed run: https://github.com/navapbc/platform-test/actions/runs/6124192950). This change applies the same fix to the deploy release step.

## Testing

Reproduced the error by running deploy-release locally
image

Fixed the error and re-ran successfully
image


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/436

PR #51, opened 2023-09-08
Refactor env configs into separate blocks

lorenyu:
## Ticket

Work for https://github.com/navapbc/template-infra/issues/425

## Changes

see title

## Context

The current app-config doesn't make it obvious how to have separate config values for each environment. This change refactors app-config to have separate config blocks for each environment which will probably be the more common scenario rather than the scenario where every environment is identically configured.

## Testing

Terraform plans for database and service layers shows no changes:

db layer:
image

service layer:
image


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/437

PR #52, opened 2023-09-12
Fix: add -reconfigure to tf init call

lorenyu:
## Ticket

n/a

## Changes

see title

## Context for reviewers

In https://github.com/navapbc/template-infra/pull/433 we switched the make infra-update-app-database and make infra-update-app-service commands to use terraform init directly rather than the terraform-init-and-apply.sh script. During that transition we didn't port over the -reconfigure option which is needed when switching around between different backend configurations (e.g. between different environments). This change fixes that.

## Testing

Ran both commands and saw that they still work and show no terraform plan changes.

update-app-service command:
image

update-app-database command:
image


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/441

PR #53, opened 2023-09-12
Fix run-database-migrations

lorenyu:
## Ticket

Work for https://github.com/navapbc/template-infra/pull/440

## Changes

see title

## Context for reviewers

In https://github.com/navapbc/template-infra/pull/440 we added an option to override task role but also changed the way the run-command script accepts environment variables, which broke migrations. This change fixes that.

## Testing

Before the fix, I ran `make release-run-database-migrations APP_NAME=app ENVIRONMENT=dev` and was able to reproduce the issue. See that environment variables don't get set:

image

After the fix, I re-ran and the migrations were successful

image

### DIFF
--- a/bin/run-database-migrations.sh
+++ b/bin/run-database-migrations.sh
@@ -58,4 +58,4 @@ ENVIRONMENT_VARIABLES=$(cat << EOF
 EOF
 )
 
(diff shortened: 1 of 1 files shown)


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/442

PR #54, opened 2023-09-12
Use migrator role when running migrations

lorenyu:
## Ticket

Work for https://github.com/navapbc/template-infra/issues/354

## Changes

see title

## Context for reviewers

Part 3 or 4 of task to separate db roles. Use the migrator role when running migrations.

## Testing

Ran db migrations locally, saw that the task role was overridden correctly

image

and the migrations completed successfully

image


charles-nava:
Screenshot of task results
Task result includes task role which is migrator as expected
Task executed successfully

PR #55, opened 2023-09-12
Remove deprecated policy and references

charles-nava:
## Ticket

Related to [#354](https://github.com/navapbc/template-infra/issues/354)

## Changes

> Database access role with broad reaching access has been deleted in favor of 2 more narrowly scoped policies
> This deletes the old policy

## Context for reviewers

> One of the last steps in the multi-stage update for #354

## Testing

- [x] Normal functionality
- [x] Run database executes successfully
- [x] Runs under correct role


charles-nava:
Screenshot 2023-09-12 at 12 03 17 PM
Most recent run of db-migrations executes successfully and under correct role

infra/modules/service/variables.tf
```diff
@@ -55,7 +55,6 @@ variable "db_vars" {
   description = "Variables for integrating the app service with a database"
   type = object({
     security_group_ids         = list(string)
-    access_policy_arn          = string
```
lorenyu, line 58: good catch. do we need to delete [this line](https://github.com/navapbc/platform-test/blob/main/infra/app/service/main.tf#L102) too

image

lorenyu (commented):
Looks good. I think I found one more line we can delete.

I think this is ready for the template-infra PR

infra/modules/service/variables.tf
charles-nava: Yes, that should be deleted. I must not have committed that change

charles-nava: Doesn't show up in [the file](https://github.com/navapbc/platform-test/blob/9f743d068626d164368c8b7170d859829c782258/infra/app/service/main.tf#L96)

charles-nava:
Handled in infra-test PR

PR #58, opened 2023-09-18
Install role manager dependencies on every apply

charles-nava:
#446

## Ticket

Resolves [#446](https://github.com/navapbc/template-infra/issues/446)

## Changes

Recursively checks for changes in vendor dir using:
`fileset(path.module, "role_manager/vendor/**")`

## Context for reviewers

Meant to prevent issue of package dependencies being missed on second engineers deployed.

## Testing

- [x] Run new fileset on reset role_manager directory confirm package installation
- [ ] Run again. Confirm no unnecessary installation
- [x] Reset directory. Run again. Confirm installation

### DIFF
--- a/infra/modules/database/role-manager.tf
+++ b/infra/modules/database/role-manager.tf
@@ -51,7 +51,7 @@ resource "aws_lambda_function" "role_manager" {
 # Installs python packages needed by the role manager lambda function before
 # creating the zip archive. Reinstalls whenever requirements.txt changes
 resource "terraform_data" "role_manager_python_vendor_packages" {
-  triggers_replace = file("${path.module}/role_manager/requirements.txt")
+  triggers_replace = fileset(path.module, "role_manager/vendor/**")
 
   provisioner "local-exec" {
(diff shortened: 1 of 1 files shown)


charles-nava:
Test 2 fails because the `triggers_replace` value is saved as empty from the previous run even though it is now full. The issue with that is if a second engineer runs with an empty vendor directory, this will register as no change and deploy the lambda without its dependencies

charles-nava:
Pivoting to run pip every time. Used `timestamp()` as `triggers_replace` value

charles-nava:
Screenshot 2023-09-18 at 11 47 34 AM
Deploy with prepared directory. Still runs pip install
Screenshot 2023-09-18 at 11 49 09 AM
Run with reset directory. Runs pip install. Packages show up in vendor dir.

lorenyu:
@charles-nava curious, if you run `make infra-update-app-database` multiple times, does the plan show a diff in the `source_code_hash` the subsequent times? I'm wondering if the zip archive remains the same, or if something about the pip install always creates a diff in the zip archive.

image

PR #62, opened 2023-10-16
Add instructions for fixing RDS issue

charles-nava:
## Ticket

Does not resolve a ticket. Documenting fix procedure for issue

## Changes

Added instructions for fixing deployment issue for Secret Manager RDS Credentials

## Testing

Procedures used by @daphnegold and @charles-nava to fix issue.


charles-nava:
Documented in release notes instead

PR #66, opened 2023-11-20
Fix "invalid count argument" error

lorenyu:
## Ticket

Work for https://github.com/navapbc/template-infra/issues/475

## Changes
* Fix "invalid count argument" error

Additional changes
* Remove "password_ts" event from role manager lambda
* Merge redundant IAM policies for role manager ssm access
* Clean up DEBUG log level configuration in role manager

## Context for reviewers

In https://github.com/navapbc/template-infra/issues/475, @rocketnova discovered a bug that prevents terraform from creating a plan for the database layer. The module sets the count for the db password secret to be `length(aws_rds_cluster.db.master_user_secret)`, but this is unnecessary since `aws_rds_cluster.db.master_user_secret` will always be available as long as the rds_cluster's `manage_master_user_password` is set to `true` which will always be the case since it is hardcoded to `true` (see https://github.com/navapbc/template-infra/blob/6b3588c9356a8134d64d145981c386df1bc09b7b/infra/modules/database/main.tf#L31).

This changeset removes the unnecessary count which fixes the terraform plan.

This changeset also includes a number of minor cleanup changes:
* Remove the "password_ts" event from the role manager lambda that was introduced in [PR 461](https://github.com/navapbc/template-infra/pull/461/files) and isn't needed.
* Merge the IAM policy that was newly created in [PR 469](https://github.com/navapbc/template-infra/pull/469/files) with the existing one that is conceptually identical.
* Clean up the DEBUG log level configuration in the role manager that was introduced in [PR 469](https://github.com/navapbc/template-infra/pull/469/files)

## Testing

1. created new workspace `lyfxcnt` ("lorenyu fix count")
   image
2. in new workspace, create db layer. screenshots of plan and results below
   image
   image
3. Created db roles with `make infra-update-app-database-roles APP_NAME=app ENVIRONMENT=dev`
   image
4. Checked roles (checking that we can connect with IAM auth and that the roles have proper permissions) with `make infra-check-app-database-roles APP_NAME=app ENVIRONMENT=dev`
   image
5. Cleaned up by remove deletion protection, deleting the db cluster and workspace
   image
   image
   image
   image
   image
   image

## Migration notes

If the rds database cluster already exists and has manage_master_user_password set to false, the terraform plan will fail with the following error:

image

thus, in order to migrate, we'll need to follow the following steps:

1. first do a targeted apply of the aws_rds_cluster by running the following command (replace ENVIRONMENT_NAME with the correct environment)

   ```
   TF_CLI_ARGS_apply='-target="module.database.aws_rds_cluster.db"' make infra-update-app-database APP_NAME=app ENVIRONMENT=<ENVIRONMENT_NAME>
   ```

   image
   image
   image

2. Then you can apply the rest of the changes normally with `make infra-update-app-database APP_NAME=app ENVIRONMENT=<ENVIRONMENT_NAME>`

   image
   image

### DIFF
--- a/infra/modules/database/role-manager.tf
+++ b/infra/modules/database/role-manager.tf
@@ -117,19 +117,7 @@ resource "aws_iam_role_policy" "ssm_access" {
         Effect   = "Allow"
         Action   = ["kms:Decrypt"]
         Resource = [data.aws_kms_key.default_ssm_key.arn]
-      }
-    ]
-  })
-}
-
-resource "aws_iam_role_policy" "database_credential_tool" {
-  count = length(aws_rds_cluster.db.master_user_secret)
-  name  = "${var.name}-role-manager-rds-ssm-access"
-  role  = aws_iam_role.role_manager.id
-
-  policy = jsonencode({
-    Version = "2012-10-17"
-    Statement = [
+      },
       {
         Effect   = "Allow"
         Action   = ["secretsmanager:GetSecretValue"]
--- a/infra/modules/database/role-manager.tf
+++ b/infra/modules/database/role-manager.tf
@@ -107,7 +107,7 @@ resource "aws_iam_role" "role_manager" {
 
 
 
-resource "aws_iam_role_policy" "ssm_access" {
+resource "aws_iam_role_policy" "role_manager_access_to_db_password" {
   name = "${var.name}-role-manager-ssm-access"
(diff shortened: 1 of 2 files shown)


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/476

PR #68, opened 2023-11-22
Enable db query logging

lorenyu:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/385

## Changes

* Attach cluster parameter group to db cluster

## Context for reviewers

We [created a cluster parameter group in the database module](https://github.com/navapbc/template-infra/blob/5dedfb0a4618a6302741ab25ed8be0d872cc075c/infra/modules/database/main.tf#L71-L87) but never attached it to the db cluster resource, so the db cluster didn't pick up the parameters. This means that query logging was never enabled. This change attaches the parameter group to the cluster which enables query logging.

## Testing

I attached the parameter group to the cluster, rebooted the instance, and ran `make infra-update-app-database-roles APP_NAME=app ENVIRONMENT=dev` and `make infra-check-app-database-roles APP_NAME=app ENVIRONMENT=dev` both of which invoke DDL commands which should be logged. Here are screenshots of those logs in AWS console:

Logs from `make infra-update-app-database-roles APP_NAME=app ENVIRONMENT=dev`
image

Logs from `make infra-check-app-database-roles APP_NAME=app ENVIRONMENT=dev`
image

### DIFF
--- a/infra/modules/database/main.tf
+++ b/infra/modules/database/main.tf
@@ -31,6 +31,8 @@ resource "aws_rds_cluster" "db" {
   storage_encrypted           = true
   kms_key_id                  = aws_kms_key.db.arn
 
+  db_cluster_parameter_group_name = aws_rds_cluster_parameter_group.rds_query_logging.name
+
   # checkov:skip=CKV_AWS_128:Auth decision needs to be ironed out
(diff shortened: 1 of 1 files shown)


lorenyu:
Done via https://github.com/navapbc/template-infra/pull/479/files

PR #70, opened 2023-12-04
Pin db cluster engine version

lorenyu:
## Ticket

Work for https://github.com/navapbc/template-infra/issues/491

## Changes

see title

## Context for reviewers

see ticket

## Testing

terraform plan shows no changes to db since the db on platform test is already on 14.6

image


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/492

PR #71, opened 2023-12-05
Check in role manager dependencies

lorenyu:
## Ticket

n/a

## Changes

* Check in role manager python dependencies
* Remove local exec pip3 install step for database module

## Context for reviewers

A number of people were confused by the fact that the terraform plan is never clean due to the triggers_replace = timestamp(). There were also concerns about the dependency on pip3. This change adds the role manager dependency to the source code, which eliminates the need for the local-exec provisioner step and also removes the dependency on pip3. The cost is that there is now some vendor code in the /modules/database/role_manager/vendor folder.

## Testing

Ran `make infra-update-app-database`
image

Ran it again to show that terraform plans now will show up as clean without the force trigger to replace
image

Ran `make infra-update-app-database-roles`
image

Ran `make infra-check-app-database-roles`
image

## Preview environment
- Service endpoint: http://p-71-app-dev-1013732362.us-east-1.elb.amazonaws.com
- Deployed commit: 38ba3b7f25eae842416b5b20a8d1ceb34ed51279


lorenyu:
Accidentally merged (my mouse was bugging out), so I reverted in #125

This was properly done as part of https://github.com/navapbc/template-infra/pull/493

PR #75, opened 2023-12-12
Add service config options to env-config

lorenyu:
## Ticket

N/A

## Changes

* Add env-config settings for service CPU, memory, and desired instance count
* Set prod to 1 vCPU, 4 GB memory, and 3 instances
* Add instructions

## Context

Add env-config settings for setting an environment's service's CPU, memory, and desired instance count. This allows a project to have separate settings for each environment.

## Testing

Temporarily set app-config's has_database to false (so we don't need to create a prod database layer), and set prod.tf's config to set network_name to `"dev"` (so we don't need to create a prod VPC), then ran terraform plan for prod. Here are the relevant sections that show that desired count is 3 and that CPU and memory settings are different from the defaults:

Ran the following commands (image_tag is the current commit on `main` branch)
```
make infra-configure-app-service APP_NAME=app ENVIRONMENT=prod
TF_CLI_ARGS_apply='-var=image_tag="0b7fbe62bb7a925ffd314dae444ab3a84b968701"' make infra-update-app-service APP_NAME=app ENVIRONMENT=prod
```
Terraform plan screenshots:

image

image

### DIFF
--- a/docs/infra/set-up-app-env.md
+++ b/docs/infra/set-up-app-env.md
@@ -10,6 +10,7 @@ Before setting up the application's environments you'll need to have:
 
 1. [A compatible application in the app folder](https://github.com/navapbc/template-infra/blob/main/template-only-docs/application-requirements.md)
 2. [Configure the app](/infra/app/app-config/main.tf). Make sure you update `has_database` to `true` or `false` depending on whether or not your application has a database to integrate with.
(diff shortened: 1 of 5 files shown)


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/503

PR #77, opened 2023-12-14
Test feature flag endpoint

lorenyu:
## Ticket

N/A

## Changes

* Hit feature-flags endpoint in example app during CI Infra tests

## Context

[This commit](https://github.com/navapbc/template-infra/commit/8bfab4387277bb7789d1aebc062bb3f6ba9a1e1d) broke the feature flag evidently integration (fixed now) but it was not caught as part of any automated testing. This change adds a ping to the feature-flags endpoint which integrates with Evidently to catch future regressions.

## Testing

Since this is just a change to CI, just leveraging CI for this


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/512

PR #80, opened 2023-12-19
Make infra service test cleanup more robust

lorenyu:
## Ticket

n/a

## Changes

see title

## Context

If there's an error in the terraform for the service layer, then the terraform apply fails, which means the cleanup step that sets force_destroy = true for the buckets will also fail since the cleanup also does a terraform apply. This change uses a targeted apply to make the cleanup step more robust. This is similar to the change in this PR — https://github.com/navapbc/template-infra/pull/487 — but for ci-infra-service.yml instead of template-only-ci-infra.yml.

## Testing

CI should cover this since this is just a change to CI


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/521

PR #83, opened 2023-12-21
Fix accounts setup for other regions

lorenyu:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/526

## Changes

see title

## Context for reviewers

@daphnegold discovered a regression where set-up-account failed when region was not us-east-1

## Testing

1. Configured my aws credentials to use us-east-2
   image

2. Configured project config's default region to us-east-2 and set project name to a new project to avoid conflict
   image

3. Ran
   ```
   make infra-set-up-account ACCOUNT_NAME=tmp
   ```
   image

4. Destroyed everything afterwards


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/534

PR #91, opened 2024-02-23
Simplify secret definitions by replace ARNs with SSM param names

lorenyu:
## Ticket

N/A

## Context for reviewers

Using the SSM param name instead of the ARN allows us to create the SSM parameter in parallel, which is useful in situations where the SSM parameter doesn't already exist before the environment is created. This is a current need on another project. 

## Testing

Spun up the service layer in a workspace `t-lyss` and checked the `/secrets` endpoint to make sure it still works

image


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/563

PR #92, opened 2024-02-26
Add app and ref to build workflow run name

lorenyu:
## Ticket

n/a

## Changes

see title

## Context for reviewers

When there are multiple apps, the build and publish workflow run name doesn't differentiate between them. This change adds the name and ref (branch or git hash) to the run name to improve developer experience.

## Testing

Run the build and publish workflow from this branch:
image


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/564

PR #94, opened 2024-02-28
Derive app_name from module path

lorenyu:
## Ticket

N/A

## Changes

see title

## Context for reviewers

When changing the name of the application from the default of `app`, such as when adding multiple applications to a monorepo, developers often forget to modify the app_name config value in the app-config module. This change removes the need for that manual step by obtaining the app_name from the name of the folder containing the application's infrastructure code.

## Testing

Running `make infra-update-app-service` yields no changes
image


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/568

PR #98, opened 2024-05-22
Move account name config into into env-config

lorenyu:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/458

## Changes

- Add shared_account_name property to app-config
- Add account_name to env-config
- Refactor account_names_by_environment
- Update docs

## Context for reviewers

For consistency with how we map environments to networks using the environment_config.network_name, this change refactors how we map environments to AWS accounts to use environment_config.account_name. There's a special case for the build-repository which is not part of any application environment but rather shared across all environments. For this we add a new property in app-config called shared_account_name. We can then delete the old account_names_by_environment map, which was a little clunky to explain.

## Testing

- Run build-and-publish workflow from this branch, which tests the `shared_account_name` app-config property by accessing `account_names_by_environment` to get the "shared" account and then logging into that account using GitHub Action's configure-aws-credentials action: https://github.com/navapbc/platform-test/actions/runs/9195210371
    image
- Run a deploy from this branch, which will log into both the "shared" account to publish the image, and into the "dev" account to deploy the image: https://github.com/navapbc/platform-test/actions/runs/9195297943
    image


lorenyu:
https://github.com/navapbc/template-infra/pull/605

PR #99, opened 2024-05-22
Fix build/publish concurrency group

lorenyu:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/606

## Changes

see title

## Context for reviewers

See context in ticket

## Testing

Run the build and publish workflow twice in quick succession, see that one is queued: 
image

See that the first one finished and successfully pushed to ECR: https://github.com/navapbc/platform-test/actions/runs/9195889400/job/25292703026
image

See that the second one sees that the image has already been published: https://github.com/navapbc/platform-test/actions/runs/9195891217/job/25292709408
image

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -27,7 +27,7 @@ jobs:
   build-and-publish:
     name: Build and publish
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/607

PR #100, opened 2024-05-26
Skip build step if image already published

lorenyu:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/610
Resolves https://github.com/navapbc/template-infra/issues/609

## Changes

- Skip build step if image already published
- Use commit hash of input.ref for build job concurrency

## Context for reviewers

The build-and-publish worklflow always builds the image and only checks if the image is already published when getting ready to publish. This change moves the published image check before the build step, allowing the workflow to skip the build step altogether if the build has already been published.

In addition, this change fixes the build job concurrency to use the commit hash of the commit being built, not the commit hash of the workflow.

## Testing

Ran build and publish workflow from this branch twice:
[1st time ran all the steps](https://github.com/navapbc/platform-test/actions/runs/9245996048)
[2nd time skipped build/publish steps](https://github.com/navapbc/platform-test/actions/runs/9245997208) and only took 10 seconds

image

To test the concurrency, I ran the job twice and showed that the second one was queued up waiting for the first to finish
image

To test that I can build two different commits simultaneously I ran it on two commits and you can see neither job is queued
- https://github.com/navapbc/platform-test/actions/runs/9246042530
- https://github.com/navapbc/platform-test/actions/runs/9246042823
image

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -43,19 +43,20 @@ jobs:
         id: check-image-published
         run: |
           IS_IMAGE_PUBLISHED=$(./bin/is-image-published.sh "${{ inputs.app_name }}" "${{ inputs.ref }}")
+          echo "Is image published: $IS_IMAGE_PUBLISHED"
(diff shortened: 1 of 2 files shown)


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/611

PR #102, opened 2024-06-05
Alter default privileges for migrator to grant all privileges to app

lorenyu:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/616

## Changes

* Alter default privileges for migrator to grant all privileges to app user on tables, sequences, and routines
* Update check function to not explicitly grant temp table access to app user
* Use print instead of logging
* Refactor lambda into separate smaller files

### Use print instead of logging library

When using logging AWS Lambda formats logs to include the timestamp, log level, and uuid, which add a lot of bytes to the log prefix. This makes the logs harder to read in the common case where we are looking at logs either in the Terminal or in GitHub Actions workflow logs. In addition, the `aws lambda invoke` CLI function only returns the last 4kb of logs for the lambda, so the addition of these log prefixes has the effect of truncating the output of the role manager function.

Also, according to [Python docs on logging](https://docs.python.org/3/howto/logging.html), in order to display console output for ordinary usage of a command line script or program, the best tool for the task is [print()](https://docs.python.org/3/library/functions.html#print). Since this is intended to be a short running program rather than something that needs to be monitored, that seems to fit the bill of using print instead of logging.

## Testing

1. Created a database in a new workspace `lydb`

2. Removed GRANT ALL from check and verified that it fails

    image

3. Ran `make infra-update-app-database-roles`
    image

4. Ran `make infra-check-app-database-roles`
    image


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/622

PR #107, opened 2024-06-20
Refine replace trigger for installation of role manager packages

lorenyu:
## Ticket

N/A

## Changes

see title

## Context for reviewers

Previously we only ran installation of role manager packages when the requirements.txt file changed, but this means that new infra engineers who ran terraform apply without having the packages installed locally would update the terraform lambda function to a zip file that didn't contain the dependencies, breaking the role manager.

In this PR https://github.com/navapbc/template-infra/pull/452/files we changed the package installation to always install, which fixed the issue but created an annoying situation where the terraform plan would always show changes to apply.

This change introduces a hybrid solution that works around the new engineer issue by triggering an installation of role manager packages locally if the zip archive doesn't exist locally. Otherwise it relies on changes to requirements.txt. It's not a perfect solution in that terraform apply needs to be run twice in order to get it into a stable state where the replacement trigger is set to the contents of requirements.txt, but it's an improvement over the previous two solutions.

## Testing

Did the following on the default workspace

Running `make infra-update-app-database` for the first time with zip file already existing
image

Running again shows clean plan
image

Removed the zip file
image

Re-ran terraform apply
image

Re-running shows a diff one more time as the triggers_replace content gets switched back from the timestamp to the file contents of the requirements.txt file
image

Finally, re-running one last time shows a clean diff
image

### DIFF
--- a/infra/modules/database/role-manager.tf
+++ b/infra/modules/database/role-manager.tf
@@ -61,7 +61,7 @@ resource "aws_lambda_function" "role_manager" {
 # environment.
 # Timestamp is used to always trigger replacement.
 resource "terraform_data" "role_manager_python_vendor_packages" {
-  triggers_replace = timestamp()
+  triggers_replace = fileexists(local.role_manager_package) ? file("${path.module}/role_manager/requirements.txt") : timestamp()
   provisioner "local-exec" {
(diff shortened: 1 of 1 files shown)


lorenyu:
done in https://github.com/navapbc/template-infra/pull/648

PR #117, opened 2024-06-26
Add check infra deploy status

lorenyu:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/623

## Changes

- Add script to check deploy status of infra changes
- Add AWS Backup to list of AWS services that GitHub Actions role needs access to

## Testing

This test run actually caught a diff in the network layer: https://github.com/navapbc/platform-test/actions/runs/9998636734

image

## Preview environment
- Service endpoint: http://p-117-app-dev-2057932227.us-east-1.elb.amazonaws.com
- Deployed commit: 52135d71e430c6a02c931e80a4b0a333c651d221


bin/infra-deploy-status-check-configs.js
```diff
@@ -0,0 +1,96 @@
...
+  if (fs.existsSync(rootModule)) {
+    return fs.readdirSync(rootModule)
+      .filter(file => file.endsWith('.s3.tfbackend'))
+      .map(file => file.replace('.s3.tfbackend', ''));
+  } else {
+    return [];
+  }
+}
+
+function getAppNames() {
```
rocketnova, line 90: This is another good reason for us to migrate to putting apps in a subdir like `/infra/apps`, so for a project with two apps, it would look something like:

```
/infra/apps/portal
/infra/apps/api
```

rocketnova:
Overall, this approach looks good to me! I think it's a positive improvement for teams. Some more thoughts inline below:

--- 

> - I added account_name to network configs, I realize networks aren't currently mapped to an account. This should probably be done in a separate PR first, which I created a ticket for:
https://github.com/navapbc/template-infra/issues/670

I like this.

> - I added the ability to log into AWS with an account name. The current custom configure-aws-credentials action allows you to log in via app_name and environment, but that doesn't apply to the account or network layer.

This is neat!

> - The database layer poses a problem due to the role manager dependencies issue, since CI will always have a fresh checkout of the repo that won't have the dependencies installed. I don't know an easy way around this so I turned off the database layer checks for now.

In the short term, I agree with your approach of excluding it. In the long term, this is another reason for us to graduate from our current role manager approach.

> - The script to generate the job matrix was complex enough that I was having a pretty hard time with it in bash, especially with the nested for loops and stuff. I'm curious what folks think about using a different language for this. I'm open to ideas here.

I think everything you have here _should_ be doable with `find`, `jq`, and string manipulation. I hesitate to introduce node into this repo. If we are going to move away from bash scripting (and I'm not sure we should, though I also have been running into frustrating limitations), then we should think about it holistically and decide what we want to move to.

lorenyu:
Done in https://github.com/navapbc/template-infra/pull/682

PR #126, opened 2024-07-18
Check in role manager archive

lorenyu:
## Ticket

## Changes

- Remove role manager dependencies from source control
- Add role manager archive to source control
- Add make command for building role manager archive
- Add documentation on how to update role manager

## Context for reviewers

The “check in role manager dependencies” change in [PR 493](https://github.com/navapbc/template-infra/pull/493) does not function as desired. The role manager lambda still shows a diff on the source_code_hash on a clean checkout of the repo. This change removes the archive_file data source altogether, and checks in the zip archive directly into the codebase. Benefits include:
- We no longer need to check in the python dependencies into source control, which is a ton of files including binary files
- There’s nothing nondeterministic for project teams, the archive file is exactly the zip package that gets deployed to lambda

The downside is that updating the role manager is more manual. So this change also adds a `make infra-module-database-role-manager-archive` to rebuild the role manager package and adds instructions for doing that.

## Testing

Emptied the vendor folder and deleted the archive, then ran

```
make infra-module-database-role-manager-archive
make infra-update-app-database APP_NAME=app ENVIRONMENT=dev
```

then tested by running

```
make infra-check-app-database-roles APP_NAME=app ENVIRONMENT=dev
```

image

## Preview environment
- Service endpoint: http://p-126-app-dev-1754676326.us-east-1.elb.amazonaws.com
- Deployed commit: 42050214f9bb1a4c815d3fbc40f5ea3916ed19e8


lorenyu:
Done in https://github.com/navapbc/template-infra/pull/695

PR #130, opened 2024-08-09
Adds scheduled jobs

coilysiren:
## Ticket

Relates to https://github.com/navapbc/template-infra/issues/525

## Changes

Adds "scheduled jobs" functionality via AWS Step Functions & EventBridge Scheduler. The step functions are configured through an array of objects inside of `infra/app/app-config/env-config/scheduled-jobs.tf`, which can configure as many step functions as one wants.

## Context for reviewers

This PR was adopted from https://github.com/HHS/simpler-grants-gov/pull/1247

## Testing

image

## Preview environment
- Service endpoint: http://p-130-app-dev-901959198.us-east-1.elb.amazonaws.com
- Deployed commit: 93423acecd595456063e9c333d12e5492d5a3a28


infra/modules/service/step-functions.tf
lorenyu, line 1: Take a pass at file names and variable names:

From [the style guide](https://github.com/navapbc/template-infra/blob/e037ced062565a484def73ba0108db6579e85a88/docs/infra/style-guide.md#L4)

> Use module names based on the logical function of the module rather than the underlying proprietary service used for implementing the module. For example, use "database" instead of "rds", or "storage" instead of "s3".

names we should change include:
- step-functions.tf
- sfn_cron_job
- etc.

infra/modules/service/step-functions.tf
```diff
@@ -0,0 +1,110 @@
...
+  definition = jsonencode({
+    "StartAt" : "ExecuteECSTask",
+    "States" : {
+      "ExecuteECSTask" : {
+        "Type" : "Task",
+        # docs: https://docs.aws.amazon.com/step-functions/latest/dg/connect-ecs.html
+        "Resource" : "arn:aws:states:::ecs:runTask.sync",
+        "Parameters" : {
+          "Cluster" : module.service.cluster_arn,
+          "TaskDefinition" : module.service.task_definition_arn,
```
lorenyu, line 37: I don't understand how this works. `module.service` isn't defined anywhere in this module.

coilysiren: Yeah thats WIP still, its a straight copy from grants.gov where the file is in a different place. Simple change to correct it, I just havent yet ^^

lorenyu (commented):
Lovely! Great work. Left some comments. Biggest one is about the design of IAM permissions. Right now you're giving a bunch of permissions to the service role and allowing the service role to pass that role on to AWS's  "states" and "scheduler" services, but that's backwards. The service role shouldn't have any of those permissions, instead the scheduler should have the policy to execute state machines and pass along the task and task_executor roles.

Also left some small comments around simplifying the design of the configuration (removing some config options) and some miscellaneous style comments.

For testing, I'd prefer to see an example that runs the app itself rather than just "ls", and to make sure the app has the proper network configuration and permissions, the app would ideally make a db query. Take a look at [the etl job](https://github.com/navapbc/platform-test/blob/1686c03c4f35d26f7ce7a4fb02d08b8e0af8cb35/app/app.py). Maybe add a new job `cron` that does something like "select 1" from the db (see the health check in the same file) and log something like "hello, world from cron" or something. And then have the test run trigger that cron job.

Non-PR feedback: For the future, to make feature branches easier to clean up, prefix your branch names with something that identifies who owns the branch e.g. `kai/sfn` instead of just `sfn`.

Thanks for all the great work!

infra/modules/service/scheduled-jobs.tf
```diff
@@ -0,0 +1,111 @@
...
+            "AwsvpcConfiguration" : {
+              "Subnets" : var.private_subnet_ids,
+              "SecurityGroups" : [aws_security_group.app.id],
+            }
+          },
+          "Overrides" : {
+            "ContainerOverrides" : [
+              {
+                "Name" : var.service_name,
+                "Environment" : each.value.environment
```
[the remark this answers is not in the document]
coilysiren: @lorenyu I'm pretty sure it's an append action, rather than an override. Although I should test that.

That aside, you don't feel like there's a high chance that people will need to use environment variables to provide some extra configuration?

infra/modules/service/scheduled-jobs.tf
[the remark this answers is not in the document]
coilysiren: This would be a great VSCode plugin for sure, although it didnt exist the last time I checked :(

infra/modules/service/scheduled-jobs.tf
```diff
@@ -0,0 +1,111 @@
...
+
+# docs: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/scheduler_schedule_group
+resource "aws_scheduler_schedule_group" "scheduled_job" {
+  for_each = {
+    for index, job in var.scheduled_jobs :
+    job.name => job
+  }
+
+  name = "${var.service_name}-${each.value.name}"
+}
```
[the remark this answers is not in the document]
coilysiren: Turns out that it's optional! I missed that on my first pass.

infra/modules/service/scheduled-jobs.tf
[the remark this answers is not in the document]
lorenyu: I'm pretty sure it's an override since we have different ENV vars per job on MA PFML.
If a project ends up thinking that they'd like different environment variables then I'd like to better understand the use case and whether they're overengineering or not. I think most common cases won't need it and it adds unnecessary complexity around testing/quality assurance to test your code with various combinations of environment variables existing or not. For example, I've seen issues where all unit tests pass and most jobs pass but some jobs fail because an environment variable was missing in that one.

In the rare case that a project truly needs it they can add the functionality themselves, but I don't really want to encourage it.

infra/modules/service/scheduled-jobs.tf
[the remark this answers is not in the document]
lorenyu: Have you tried this one? Looks like it might work: https://marketplace.visualstudio.com/items?itemName=TylerHarris.terraform-link-docs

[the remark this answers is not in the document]
lorenyu: Just tried it, seems to work so far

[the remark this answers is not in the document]
coilysiren: Ahah, nice!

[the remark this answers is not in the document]
coilysiren: I feel like the last time I looked for this must have been 2020 or something, heh. Regardless, glad that it exists now

lorenyu:
@coilysiren since i'll be out friday, just one more thing if you end up having time for it on friday:
as part of this PR can you also [update the docs on background jobs](https://github.com/navapbc/platform-test/blob/b393f151b27a56ad34d5d34094c5b4a9dc970b5c/docs/infra/background-jobs.md)

infra/project-config/aws-services.tf
```diff
@@ -19,6 +19,7 @@ locals {
     "kms",
     "lambda",
     "logs",
+    "mobiletargeting",
```
coilysiren, line 22: Not sure if I should leave this in?

app/app.py
```diff
@@ -63,10 +63,7 @@ def document_upload():
     path = f"uploads/{datetime.now().date()}/${{filename}}"
     upload_url, fields = storage.create_upload_url(path)
     additional_fields = "".join(
-        [
-            f'<input type="hidden" name="{name}" value="{value}">'
-            for name, value in fields.items()
-        ]
+        [f'<input type="hidden" name="{name}" value="{value}">' for name, value in fields.items()]
```
coilysiren, line 66: linter did this, I'll remove it in a bit

infra/modules/service/variables.tf
```diff
@@ -56,6 +56,11 @@ variable "enable_command_execution" {
   description = "Whether the service should enable ECS Exec, such as for debugging"
 }
 
+variable "environment_name" {
+  type        = string
+  description = "name of the application environment"
+}
+
```
lorenyu, line 63: see earlier comments i think we can get rid of this

infra/modules/service/scheduled_jobs_scheduler_role.tf
lorenyu, line 1: /i appreciate the descriptiveness of "scheduled jobs scheduler role.tf" but in this case i think "scheduler_role.tf" is sufficient

infra/project-config/aws-services.tf
lorenyu: i forget why we needed it

lorenyu (commented):
looks great. no major feedback, just left some polish nits.

infra/app/app-config/env-config/scheduled_jobs.tf
```diff
@@ -0,0 +1,12 @@
+locals {
+  # The `task_command` is what you want your scheduled job to run, for example: ["poetry", "run", "flask"].
+  # Schedule expression defines the frequency at which the job should run.
+  # The syntax for `schedule_expression` is explained in the following documentation:
+  # https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-scheduled-rule-pattern.html
+  scheduled_jobs = {
+    show-files = {
```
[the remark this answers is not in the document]
coilysiren: Oops, good catch

infra/modules/service/scheduled_jobs_scheduler_role.tf
```diff
@@ -0,0 +1,73 @@
...
+      identifiers = [
+        "scheduler.amazonaws.com"
+      ]
+    }
+  }
+}
+
+data "aws_iam_policy_document" "scheduled_jobs_scheduler_policy" {
+
+  # policy sourced via: https://docs.aws.amazon.com/step-functions/latest/dg/stepfunctions-iam.html
```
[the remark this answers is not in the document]
coilysiren: OH NO!!! AWS killed my link 😭

infra/modules/service/scheduled_jobs_workflow_role.tf
```diff
@@ -0,0 +1,108 @@
...
+
+  # policy sourced via: https://docs.aws.amazon.com/step-functions/latest/dg/ecs-iam.html
+  statement {
+    sid = "StepFunctionsRunTask"
+    actions = [
+      "ecs:RunTask",
+      "ecs:StopTask",
+      "ecs:DescribeTasks",
+    ]
+    resources = ["*"]
```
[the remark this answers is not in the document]
coilysiren: Unfortunately not, because this is working on tasks, where the tasks aren't defined at build time. This in contrast to the cluster, which is defined at build time. Also the arn for the tasks is a UUID. At best, we could do something like:

```
condition
  type = StringLike
  key = tags!ecs-cluster
  value = aws_ecs_cluster.cluster.name
```

Which would of course require us to tag our ECS tasks.

So its possible to scope this down further, but I wouldn't really recommend it.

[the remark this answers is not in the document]
lorenyu: Hmm I'm skeptical. If you look at events_role.tf it's working on tasks:
```
data "aws_iam_policy_document" "run_task" {
  statement {
    effect    = "Allow"
    actions   = ["ecs:RunTask"]
    resources = ["${aws_ecs_task_definition.app.arn_without_revision}:*"]
    condition {
      test     = "ArnLike"
      variable = "ecs:cluster"
      values   = [aws_ecs_cluster.cluster.arn]
    }
  }
```
Seems like if there are multiple apps, the workflow orchestrator for one app shouldn't be able to run tasks for the other app

infra/modules/service/scheduled_jobs_scheduler_role.tf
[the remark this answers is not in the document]
coilysiren: remove the links since they're broken now ☹️

(I mean I could have found the new links, but chose not to)

docs/infra/background-jobs.md
```diff
@@ -2,7 +2,7 @@
 
 The application may have background jobs that support the application. Types of background jobs include:
 
-* Jobs that occur on a fixed schedule (e.g. every hour or every night) — This type of job is useful for ETL jobs that can't be event-driven, such as ETL jobs that ingest source files from an SFTP server or from an S3 bucket managed by another team that we have little control or influence over. **This functionality has not yet been implemented**
```
coilysiren, line 5: 🎉

infra/modules/service/workflow_orchestrator_role.tf
```diff
@@ -0,0 +1,110 @@
...
+
+resource "aws_iam_role" "workflow_orchestrator" {
+  name                = "${var.service_name}-workflow-orchestrator"
+  managed_policy_arns = [aws_iam_policy.workflow_orchestrator.arn]
+  assume_role_policy  = data.aws_iam_policy_document.workflow_orchestrator_assume_role.json
+}
+
+data "aws_iam_policy_document" "workflow_orchestrator_assume_role" {
+  statement {
+    sid     = "ECSTasksAssumeRole"
```
lorenyu, line 14: Looks like this name was copy pasted, rename or maybe we don't even need a SID since there's only one statement

infra/modules/service/workflow_orchestrator_role.tf
```diff
@@ -0,0 +1,110 @@
...
+    actions = ["sts:AssumeRole"]
+    principals {
+      type        = "Service"
+      identifiers = ["states.amazonaws.com"]
+    }
+    condition {
+      test     = "ArnLike"
+      variable = "aws:SourceArn"
+      values   = ["arn:aws:states:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:stateMachine:*"]
```
lorenyu, line 23: we actually have the state machine ARNs available so instead of "*" let's use those:
```suggestion
      values   = [aws_sfn_state_machine.scheduled_jobs[*].arn]
```

lorenyu (commented):
Approved! Just requesting one final IAM tweak, then after you test that it works we can create the PR in template-infra and we're good to go.

infra/modules/service/workflow_orchestrator_role.tf
coilysiren: The state machine itself uses this role, so unfortunately its a circular dependency.

```
terraform apply -var "environment_name=dev"
╷
│ Error: Cycle: module.service.data.aws_iam_policy_document.workflow_orchestrator_assume_role, module.service.aws_iam_role.workflow_orchestrator, module.service.aws_sfn_state_machine.scheduled_jobs
```

infra/modules/service/workflow_orchestrator_role.tf
coilysiren: Updated! https://github.com/navapbc/platform-test/pull/130/commits/367badc4c79286c9215692073b227ac4234d6446

infra/modules/service/workflow_orchestrator_role.tf
lorenyu: Oh ok can we manually construct the ARN then:

```suggestion
      values   = [for job_name in keys(var.scheduled_jobs) "arn:aws:states:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:stateMachine:${var.service_name}-${job_name}"]
```

coilysiren: Unfortunately I can't get any get this to work. The error message I'm getting is:

```
The principal states.amazonaws.com is not authorized to assume the provided role. (role: arn:aws:iam::{REDACTED account id}:role/p-130-app-dev-workflow-orchestrator)
```

Which doesn't make sense to me because the arn is an exact match 🤷🏼

lorenyu: Alright no big deal we can always refine in the future. Thanks!

PR #135, opened 2024-09-03
Only run PR env checks if PR open

lorenyu:
## Ticket

N/A

## Changes

see title

## Context for reviewers

I believe there's a race condition if you push a change to a PR then close it soon after, the destroy PR environment workflow might run before the workflow that creates/updates the PR environment.

## Testing

I don't really know how to reproduce this, but this seems like an appropriate change to make.

PR environments still work as you can see on this PR:

## Preview environment
- Service endpoint: http://p-135-app-dev-1154552620.us-east-1.elb.amazonaws.com
- Deployed commit: 95c9440dccf839b1011f91fd35c5b77693757315


lorenyu:
Pushed this to template-infra main directly by accident

PR #136, opened 2024-09-13
Run s3 jobs via step functions

coilysiren:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/744

## Changes

This PR changes the Event Bridge file upload job from triggering an ECS task directly, to triggering a step function that then triggers an ECS task. We are doing this because Step Functions brings a basic observability layer (it shows failures and successes) to the triggering of ECS tasks.

## Context for reviewers

This PR looks a lot like [the scheduled jobs PR](https://github.com/navapbc/platform-test/pull/130). It even shares a role with the scheduled jobs (`infra/modules/service/workflow_orchestrator_role.tf`)

## Testing

_(the gif runs through the whole process end to end, and ends up 70 seconds long)_

https://github.com/user-attachments/assets/552cc075-6332-4586-b930-640d0f4993f8

## Preview environment
- Service endpoint: http://p-136-app-dev-1918576231.us-east-1.elb.amazonaws.com
- Deployed commit: a2edc4ddb2d6db8a5341125955138d55e233f47e


infra/modules/service/events_role.tf
```diff
@@ -30,27 +30,35 @@ resource "aws_iam_policy" "run_task" {
 
 data "aws_iam_policy_document" "run_task" {
```
coilysiren, line 31: This file looks *exactly* like the one for the scheduler role:

https://github.com/navapbc/platform-test/blob/main/infra/modules/service/scheduler_role.tf

But it didn't seem right to share them, given that they are acting on different arns, and are assumed by different services.

infra/modules/service/events_jobs.tf
```diff
@@ -98,10 +86,70 @@ resource "aws_cloudwatch_event_target" "document_upload_jobs" {
...
       ]
     }), "\\u003c", "<"), "\\u003e", ">")
   }
 }
+
+resource "aws_sfn_state_machine" "file_upload_jobs" {
+  for_each = var.file_upload_jobs
+
+  name     = "${var.service_name}-${each.key}"
+  role_arn = aws_iam_role.workflow_orchestrator.arn
```
coilysiren, line 100: This is the same role that the cron scheduler uses 😎

infra/modules/service/events_jobs.tf
```diff
@@ -98,10 +86,85 @@ resource "aws_cloudwatch_event_target" "document_upload_jobs" {
     input_template = replace(replace(jsonencode({
       containerOverrides = [
         {
-          name    = local.container_name,
           command = each.value.task_command
         }
       ]
     }), "\\u003c", "<"), "\\u003e", ">")
   }
 }
```
lorenyu, line 94: I think we can simplify this:

```
input_template = replace(replace(jsonencode({
  task_command = each.value.task_command
}), "\\u003c", "<"), "\\u003e", ">")
```

then below where we set the container overrides we can do

```
"Command.$" : "$.task_command"
```

infra/modules/service/events_jobs.tf
```diff
@@ -98,10 +86,85 @@ resource "aws_cloudwatch_event_target" "document_upload_jobs" {
...
+                #         "<task_command_arg_2>"
+                #         ...
+                #       ]
+                #     }
+                #   ]
+                # }
+                #
+                # The syntax for parsing the input data comes from JSONPath.
+                "Name" : local.container_name,
+                "Command.$" : "$.containerOverrides[0].command[*]"
```
lorenyu, line 138: See my previous comment but I think we can simplify the input transformer and this line.

infra/modules/service/events_jobs.tf
```diff
@@ -98,10 +86,85 @@ resource "aws_cloudwatch_event_target" "document_upload_jobs" {
...
+
+  tracing_configuration {
+    enabled = true
+  }
+}
+
+resource "aws_cloudwatch_log_group" "file_upload_jobs" {
+  for_each = var.file_upload_jobs
+
+  name_prefix = "/aws/vendedlogs/states/${var.service_name}-${each.key}"
```
lorenyu, line 162: I just realized that if someone names one of their file_upload_jobs the same as one of their scheduled_jobs then the logs would conflict. Could we update this to:
```suggestion
  name_prefix = "/aws/vendedlogs/states/file-upload-jobs/${var.service_name}-${each.key}"
```

and also make a similar change in scheduled_jobs.tf to add `/scheduled-jobs/` into the path

infra/modules/service/events_role.tf
```diff
@@ -30,27 +30,35 @@ resource "aws_iam_policy" "run_task" {
...
+  }
+
+  dynamic "statement" {
+    for_each = aws_sfn_state_machine.file_upload_jobs
+
+    content {
+      actions = [
+        "states:StartExecution",
+      ]
+      resources = [statement.value.arn]
```
lorenyu, line 49: I didn't realize this previously when we worked on the scheduled jobs, but wouldn't this be much simpler as a non-dynamic statement with an array for the resources list:

```suggestion
  statement {
    actions = [
      "states:StartExecution",
    ]
    resources = [for job in aws_sfn_state_machine.file_upload_jobs : job.arn]
```

can we try this? If it works can we make the same change to scheduler_role.tf?

infra/modules/service/events_role.tf
```diff
@@ -30,27 +30,35 @@ resource "aws_iam_policy" "run_task" {
...
-      values   = ["ecs-tasks.amazonaws.com"]
+  dynamic "statement" {
+    for_each = aws_sfn_state_machine.file_upload_jobs
+
+    content {
+      actions = [
+        "states:DescribeExecution",
+        "states:StopExecution",
+      ]
+      resources = ["${statement.value.arn}:*"]
```
lorenyu, line 61: Similar to my previous comment, would this be simpler as
```suggestion
  statement {
    actions = [
      "states:DescribeExecution",
      "states:StopExecution",
    ]
    resources = [for job in aws_sfn_state_machine.file_upload_jobs : "${job.arn}:*"]
```

lorenyu (approved):
I left some comments about simplifying a few things but otherwise looks amazing.

lorenyu:
Not important but curious why you closed this

coilysiren:
Yeah I dunno, shouldn't have closed this one before https://github.com/navapbc/template-infra/pull/757 was merged

PR #137, opened 2024-09-18
Add input for which version to deploy on `Deploy App` workflow

doshitan:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/693

## Changes

Pass the value along to the other workflows as necessary.

## Testing

Deploying with defaults (`dev` env and `main` version): https://github.com/navapbc/platform-test/actions/runs/10926380147

~Deploying this branch (which is functionally same as `main` at time of testing) to `dev` to see the different version get carried through and succeed: https://github.com/navapbc/platform-test/actions/runs/10926808599~ outdated

~Deploying this branch (which is functionally same as `main` at time of testing) to `dev` to see the different version get carried through and succeed
https://github.com/navapbc/platform-test/actions/runs/11000627559~ outdated

Deploying this branch (which is functionally same as `main` at time of testing) to `dev` to see the different version get carried through and succeed: https://github.com/navapbc/platform-test/actions/runs/11040058211

## Preview environment
- Service endpoint: http://p-137-app-dev-415399966.us-east-1.elb.amazonaws.com
- Deployed commit: 54559c3256065026290b80dd051aba6ed5e042fc

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -12,6 +12,10 @@ on:
         description: The branch, tag or SHA to checkout. When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event. Otherwise, use branch or tag that triggered the workflow run.
         required: true
         type: string
+    outputs:
+      commit_hash:
+        description: The SHA that was built
+        value: ${{ jobs.get-commit-hash.outputs.commit_hash }}
   workflow_dispatch:
     inputs:
       app_name:
@@ -63,7 +67,7 @@ jobs:
       - name: Check if image is already published
         id: check-image-published
         run: |
-          is_image_published=$(./bin/is-image-published "${{ inputs.app_name }}" "${{ inputs.ref }}")
+          is_image_published=$(./bin/is-image-published "${{ inputs.app_name }}" "${{ needs.get-commit-hash.outputs.commit_hash }}")
           echo "Is image published: $is_image_published"
           echo "is_image_published=$is_image_published" >> "$GITHUB_OUTPUT"
 
--- a/.github/workflows/cd-app.yml
+++ b/.github/workflows/cd-app.yml
@@ -14,7 +14,7 @@ on:
   workflow_dispatch:
     inputs:
       environment:
-        description: Environment to be deployed to
+        description: Environment to deploy to
         required: true
         default: "dev"
         type: choice
@@ -25,7 +25,7 @@ on:
       version:
         required: true
         default: "main"
-        description: Tag or branch or SHA to be deployed
+        description: Tag or branch or SHA to deploy
 
 jobs:
   deploy:
--- a/.github/workflows/cd-app.yml
+++ b/.github/workflows/cd-app.yml
@@ -1,7 +1,7 @@
 name: Deploy App
 # Need to set a default value for when the workflow is triggered from a git push
 # which bypasses the default configuration for inputs
-run-name: Deploy ${{ github.ref_name }} to App ${{ inputs.environment || 'dev' }}
+run-name: Deploy ${{ inputs.version || 'main' }} to App ${{ inputs.environment || 'dev' }}
 
 on:
   push:
@@ -14,7 +14,7 @@ on:
   workflow_dispatch:
     inputs:
       environment:
-        description: "target environment"
+        description: Environment to be deployed to
         required: true
         default: "dev"
         type: choice
@@ -22,6 +22,10 @@ on:
           - dev
           - staging
           - prod
+      version:
+        required: true
+        default: "main"
+        description: Tag or branch or SHA to be deployed
 
 jobs:
   deploy:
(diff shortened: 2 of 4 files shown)


doshitan:
Markdown lint failure unrelated to these changes, created https://github.com/navapbc/template-infra/issues/750

.github/workflows/cd-app.yml
```diff
@@ -14,14 +14,18 @@ on:
...
-        description: "target environment"
+        description: Environment to be deployed to
         required: true
         default: "dev"
         type: choice
         options:
           - dev
           - staging
           - prod
+      version:
```
doshitan, line 25: Was between `ref` and `version`. Went with `version` as I think this workflow is a little more non-technical user facing, and for the odd chance "versions" in some project may not correspond to something directly in git (would be pretty weird) and so they might want to update this to do a little translation before passing on to the actual deploy job. But no strong opinion here.

I do wonder if the other inputs added to the "lower level" workflows should maybe use `ref`? Like we do in build-and-publish?

And related side note, I noticed some workflows use `commit_hash` as input name for largely the same purpose. Should we standardize on one naming convention for "git reference thing"? Or is the naming there intentional that those other workflows specifically require a full commit SHA?

.github/workflows/database-migrations.yml
```diff
@@ -32,6 +36,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.version }}
```
lorenyu, line 40: Rather than checking out inputs.version, we want instead to pass in IMAGE_TAG to `make release-run-database-migrations`

.github/workflows/deploy.yml
```diff
@@ -32,6 +37,8 @@ jobs:
       id-token: write
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.version }}
```
lorenyu, line 41: Similar to previous comment, rather than checking out inputs.version, we want instead to pass in IMAGE_TAG to `make release-deploy`

lorenyu (commented):
Looks good, just one important change to pass in IMAGE_TAG rather than checking out a different ref

.github/workflows/database-migrations.yml
doshitan: I was a little worried about backwards incompatibilities, but I guess you're running the workflow from `main` anyway (reflecting current deploy processes), and if something about migrations/deploying the app has changed in an incompatible way with older commits, you'll have to manually intervene anyway.

So can update for this.

.github/workflows/cd-app.yml
lorenyu: Sounds good with the choice to use `version`. Agree that having a name that reflects semantics of the release process rather than the underlying implementation (git refs) makes sense here.

deploy.yml and database-migrations.yml are callable workflows, but in my mind they are still relatively high level workflows, whereas build-and-publish.yml is lower level. Somewhat subjective but I'm fine keeping it the way you have it already here.

re: flows that use `commit_hash`, yeah those flows require the full Sha. A ref won't work. The commit_hash is used directly to set the terraform variable image_tag in the terraform apply command. That flow explicitly doesn't reuse the deploy workflow because we want to bypass running database migrations, which would be hairy trying to support on PR environments since all PR environments share the database with the dev environment.

lorenyu (approved):
LGTM, feel free to create the template-infra PR now.
If you are inclined to you can test a situation where the branch we're running the workflow from is different from the version we're deploying. E.g. you can make a change to the test app in this branch (e.g. change the hello world string), then undo the change, then run a deploy from `main` branch but pass in the commit SHA of the commit that had the change and deploy to `dev`.

But I don't feel strongly about that level of test, this feels like a reasonably low risk change to me.

doshitan:
> Looks good, just one important change to pass in IMAGE_TAG rather than checking out a different ref

Going over it again, I don't think we want to just pass in `version` as the `IMAGE_TAG` since I think we'd want the same branch, deployed multiple times, to always deploy that branches top commit right? So we need to resolve the `version` to the actual commit and have all the things that build and deploy the image use that for `IMAGE_TAG`.

.github/workflows/build-and-publish.yml
```diff
@@ -63,7 +67,7 @@ jobs:
       - name: Check if image is already published
         id: check-image-published
         run: |
-          is_image_published=$(./bin/is-image-published "${{ inputs.app_name }}" "${{ inputs.ref }}")
+          is_image_published=$(./bin/is-image-published "${{ inputs.app_name }}" "${{ needs.get-commit-hash.outputs.commit_hash }}")
```
lorenyu, line 70: Thank you for finding and fixing this bug! I could have sworn I fixed this issue but I was still seeing it happen but didn't have time to look into it

lorenyu (approved):
Good point on needing to get the commit hash not just the version. Maybe test it using this branch to make sure it works?

doshitan:
Merged in https://github.com/navapbc/template-infra/commit/4275fd01232d8563491d478aebfef8d14874469b

PR #138, opened 2024-09-24
Fix step functions diffs

coilysiren:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/751

## Changes

Removes a dynamic block, in favor of an inline for function.

## Context for reviewers

[An example of the problem this solves is here.](https://github.com/navapbc/platform-test/actions/runs/11008727995/job/30566934494)

## Testing

Here's the diff against main:

```
terraform apply -var "environment_name=dev"

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # module.service.aws_iam_policy.scheduler will be updated in-place
  ~ resource "aws_iam_policy" "scheduler" {
        id               = "arn:aws:iam::430004246987:policy/app-dev-scheduler"
        name             = "app-dev-scheduler"
      ~ policy           = jsonencode(
          ~ {
              ~ Statement = [
                    {
                        Action   = [
                            "events:PutTargets",
                            "events:PutRule",
                            "events:DescribeRule",
                        ]
                        Effect   = "Allow"
                        Resource = "arn:aws:events:us-east-1:430004246987:rule/StepFunctionsGetEventsForStepFunctionsExecutionRule"
                        Sid      = "StepFunctionsEvents"
                    },
                  ~ {
                      + Sid      = "StepFunctionsExecution"
                        # (3 unchanged attributes hidden)
                    },
                  ~ {
                      + Sid      = "StepFunctionsDescribeStop"
                        # (3 unchanged attributes hidden)
                    },
                ]
                # (1 unchanged attribute hidden)
            }
        )
        tags             = {}
        # (7 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

You can see here that the only change is the Sids I added. That should be the same change once we merge this, updating the Sids.

## Preview environment
- Service endpoint: http://p-138-app-dev-122029000.us-east-1.elb.amazonaws.com
- Deployed commit: 4640ac8189b816b16acb6e556ce3b9769f0d6aa1

### DIFF
--- a/infra/modules/service/scheduler_role.tf
+++ b/infra/modules/service/scheduler_role.tf
@@ -39,7 +39,7 @@ data "aws_iam_policy_document" "scheduler" {
   statement {
     sid = "StepFunctionsExecution"
     actions = [
-      "states:StartExecution",
+      "states:StartExecution"
     ]
     resources = [for job in keys(var.scheduled_jobs) : "arn:aws:states:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:stateMachine:${var.service_name}-${job}"]
   }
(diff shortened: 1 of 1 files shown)


lorenyu (approved):

lorenyu (commented):
Actually, is this even an issue? This seems to make things less DRY, because if we change the way the state machine is named then we'd have to change it here too

PR #140, opened 2024-10-04
Email Notifications [mimics main branch]

coilysiren:
## Ticket

See also: https://github.com/navapbc/platform-test/pull/141

Resolves https://github.com/navapbc/template-infra/issues/690

## Context

See https://github.com/navapbc/platform-test/pull/141 for a comprehensive rundown of the changes. [This PR includes a hardcoded exception](https://github.com/navapbc/platform-test/pull/140/commits/d8ecbfe8c48498bd316c5813cd85f13598e5aa04) to make this PR branch mimic a main branch. That is, to make it deploy the Email Notifications resources.

## Testing

image

## Preview environment
- Service endpoint: http://p-140-app-dev-1104456610.us-east-1.elb.amazonaws.com
- Deployed commit: e82004499fb579878d64f6931c81572444701042

PR #141, opened 2024-10-26
Kai/notifications 2

coilysiren:
## Ticket

## Changes

## Context for reviewers

## Testing


infra/app/service/identity_provider.tf
```diff
@@ -0,0 +1,54 @@
+locals {
```
coilysiren, line 1: This change (eg. the whole file) is tangential to the main point of the PR

infra/modules/notifications/resources/logs.tf
```diff
@@ -0,0 +1,23 @@
...
+  matching_types = [
+    "bounce",
+    "click",
+    "complaint",
+    "delivery",
+    "open",
+    "reject",
+    "renderingFailure",
+    "send"
+  ]
```
coilysiren, line 16: I haven't taken the time to figure out what adding to this list means, so I just added all the options

infra/modules/identity-provider/resources/main.tf
```diff
@@ -34,12 +29,12 @@ resource "aws_cognito_user_pool" "main" {
...
-    email_sending_account = var.sender_email != null ? "DEVELOPER" : "COGNITO_DEFAULT"
+    source_arn            = var.email_identity_arn
+    email_sending_account = var.email_identity_arn != null ? "DEVELOPER" : "COGNITO_DEFAULT"
     # Customize the name that users see in the "From" section of their inbox, so that it's clearer who the email is from.
     # This name also needs to be updated manually in the Cognito console for each environment's Advanced Security emails.
-    from_email_address     = var.sender_email != null ? (var.sender_display_name != null ? "${var.sender_display_name} <${var.sender_email}>" : var.sender_email) : null
-    reply_to_email_address = var.reply_to_email != null ? var.reply_to_email : null
+    from_email_address     = var.email_identity_arn != null ? (var.sender_display_name != null ? "${var.sender_display_name} <${var.sender_email}>" : var.sender_email) : null
+    reply_to_email_address = var.reply_to_email
   }
```
coilysiren, line 38: Rocket made these changes, but they LGTM. They change the IDP to use the same email as SES / Pinpoint

infra/app/service/notifications.tf
```diff
@@ -0,0 +1,50 @@
...
+  email_identity_arn = module.app_config.enable_notifications ? (
+    !local.is_temporary ? module.notifications[0].email_identity_arn : module.existing_notifications[0].email_identity_arn
+  ) : null
+  email_identity_config = module.app_config.enable_notifications ? (
+    !local.is_temporary ? module.notifications[0].email_identity_config : module.existing_notifications[0].email_identity_config
+  ) : null
+  notifications_environment_variables = module.app_config.enable_notifications ? {
+    PINPOINT_APP_ID = module.notifications_app[0].app_id,
+    PINPOINT_SENDER = local.notifications_config.sender_email
+  } : {}
```
coilysiren, line 12: [via AWS's example code](https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/python/example_code/pinpoint/pinpoint_send_email_message_api.py#L77-L80), these should be the only two things you need to use the Pinpoint API inside of an application

infra/project-config/aws_services.tf
```diff
@@ -19,6 +19,7 @@ locals {
     "kms",
     "lambda",
     "logs",
+    "mobiletargeting",
```
coilysiren, line 22: mobiletargeting == Pinpoint

infra/app/service/identity_provider.tf
lorenyu: wanna just move this change to a separate PR that we can merge quickly? seems like an easy one to pull out since it should just result in no-op terraform plans

infra/modules/notifications/resources/main.tf
```diff
@@ -0,0 +1,77 @@
...
+      Version = "2008-10-17",
+      Statement = [
+        {
+          Sid    = "PinpointEmail",
+          Effect = "Allow",
+          Principal = {
+            Service = "pinpoint.amazonaws.com"
+          },
+          Action   = "ses:*",
+          Resource = "${aws_sesv2_email_identity.sender.arn}",
```
lorenyu, line 57: nit don't think you need the `${}`
```suggestion
          Resource = aws_sesv2_email_identity.sender.arn,
```

infra/modules/notifications/resources/main.tf
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  suppression_options {
+    suppressed_reasons = ["BOUNCE", "COMPLAINT"]
+  }
+}
+
+# Allow AWS Pinpoint to send email on behalf of this email identity.
+# Docs: https://docs.aws.amazon.com/pinpoint/latest/developerguide/security_iam_id-based-policy-examples.html#security_iam_resource-based-policy-examples-access-ses-identities
+resource "aws_sesv2_email_identity_policy" "sender" {
```
lorenyu, line 42: nit: can we move this IAM policy to it's own file like `access-control.tf`

infra/modules/notifications/resources/main.tf
```diff
@@ -0,0 +1,77 @@
...
+
+# Verify email sender identity.
+# Docs: https://docs.aws.amazon.com/pinpoint/latest/userguide/channels-email-manage-verify.html
+resource "aws_sesv2_email_identity" "sender" {
+  email_identity         = local.stripped_domain_name
+  configuration_set_name = aws_sesv2_configuration_set.email.configuration_set_name
+}
+
+# The configuration set applied to messages that is sent through this email channel.
+resource "aws_sesv2_configuration_set" "email" {
```
lorenyu, line 20: I see you attach this configuration set as the default configuration set for the email identity, and also set this same configuration set as the default configuration set for the email channel you create in notifications-app. I think we should only do one of the two, not both as it's redundant. Perhaps we could discuss which would be more appropriate to do.

My initial thoughts are that I lean towards starting with the simplest option:
1. Set the default configuration set for the email identity as you already do in this file
2. Don't set any default configuration set for the email channel in modules/notifications-app/resources/main.tf

Once we do that we can simplify the interfaces a bit too and remove the configuration set from the outputs of the module/notifications module (the module that creates the email identity) and also remove it from the variables of the notifications-app module.

infra/modules/notifications/resources/main.tf
lorenyu, line 1: nit: i think it's more intuitive to rename this module `notifications-email-identity` (instead of `notifications`) and then rename `notifications-app` to `notifications`

infra/modules/notifications-app/resources/main.tf
```diff
@@ -0,0 +1,10 @@
+resource "aws_pinpoint_app" "app" {
+  name = var.name
+}
+
+resource "aws_pinpoint_email_channel" "app" {
```
lorenyu, line 5: nit: can we move this resource to its own file? something like:
- `email.tf`
- `email-notifications.tf`
- `email-channel.tf`

we eventually also want to add an SMS channel so might as well organize things now

infra/modules/notifications-app/resources/main.tf
```diff
@@ -0,0 +1,10 @@
...
+  application_id    = aws_pinpoint_app.app.application_id
+  configuration_set = var.email_identity_config
```
lorenyu, line 7: see my other comment, i think we can get rid of this option and also get rid of the var.email_identity_config variable

infra/modules/notifications-app/resources/variables.tf
```diff
@@ -0,0 +1,25 @@
+variable "name" {
+  type        = string
+  description = "Name of the notifications project/application"
+}
+
+variable "sender_email" {
+  type        = string
+  description = "Email address to use to send notification emails"
+}
```
lorenyu, line 9: couldn't we get the sender_email by using a data source with the email_identity_arn and doing data.email_identity_arn.sender.email? that would simplify the variables a bit

infra/modules/notifications-app/resources/variables.tf
```diff
@@ -0,0 +1,25 @@
...
+
+variable "email_identity_arn" {
+  type        = string
+  description = "The ARN of the email identity to use for sending emails"
+}
+
+variable "email_identity_config" {
+  type        = string
+  description = "The name of the email configuration set to use for sending emails"
+}
```
lorenyu, line 25: see other comment, i think we can get rid of this

infra/modules/notifications/data/outputs.tf
```diff
@@ -0,0 +1,7 @@
+output "email_identity_config" {
+  value = data.aws_sesv2_configuration_set.main.configuration_set_name
+}
```
lorenyu, line 3: see other comment i think we can get rid of this

infra/modules/notifications/resources/outputs.tf
```diff
@@ -0,0 +1,7 @@
+output "email_identity_config" {
+  value = aws_sesv2_configuration_set.email.configuration_set_name
+}
```
lorenyu, line 3: see other comment i think we can get rid of this

infra/project-config/aws_services.tf
lorenyu: thanks, can we add it as an inline comment

infra/app/service/notifications.tf
```diff
@@ -0,0 +1,50 @@
+locals {
+  # If this is a temporary environment, re-use an existing email identity. Otherwise, create a new one.
+  email_identity_arn = module.app_config.enable_notifications ? (
+    !local.is_temporary ? module.notifications[0].email_identity_arn : module.existing_notifications[0].email_identity_arn
+  ) : null
+  email_identity_config = module.app_config.enable_notifications ? (
```
lorenyu, line 6: see other comment i think we can get rid of this

infra/modules/notifications/resources/variables.tf
```diff
@@ -0,0 +1,20 @@
...
+
+variable "name" {
+  type        = string
+  description = "Name of the notifications project/application"
+}
+
+variable "sender_email" {
+  type        = string
+  description = "Email address to use to send notification emails"
+}
```
lorenyu, line 14: nice to have, can we add a [validation rule](https://developer.hashicorp.com/terraform/language/values/variables#custom-validation-rules) that sender email is a string that ends with `"@${var.domain_name}"`

lorenyu (commented):
sorry almost done reviewing but still looking through a few things

infra/modules/notifications/resources/main.tf
lorenyu: After re-reading this PR, I realize this is creating a domain identity not an email identity, so I changed my mind and think we should name this module something like `notifications-domain` or `notifications-verified-domain` or `notifications-sender-domain`.

And we should rename variables like "email_identity" to "domain_identity", and resources like "aws_sesv2_email_identity.sender" to "aws_sesv2_email_identity.sender_domain"

infra/modules/notifications/data/main.tf
```diff
@@ -0,0 +1,11 @@
+locals {
+  stripped_domain_name = replace(var.domain_name, "/[.]$/", "")
```
lorenyu, line 2: I see this stripping of the domain name a number of times in this PR, but don't understand why it's necessary. service_config.domain_name doesn't have a trailing period.

infra/modules/notifications/resources/variables.tf
```diff
@@ -0,0 +1,20 @@
+variable "domain_name" {
+  description = "The domain name to configure SES."
+  type        = string
+}
+
+variable "name" {
+  type        = string
+  description = "Name of the notifications project/application"
+}
```
lorenyu, line 9: Get rid of one of these variables to keep the interface simple and also so that we can have the interface to the data module be a single variable. I lean towards getting rid of `name`. `domain_name` is more explicit which I think is good. Resources that rely on `name` for naming thing like the logs resource can use `domain_name` instead.

infra/modules/notifications/data/variables.tf
```diff
@@ -0,0 +1,9 @@
+variable "name" {
+  type        = string
+  description = "Name of the notifications project/application"
+}
+
+variable "domain_name" {
+  type        = string
+  description = "The domain name to use for the email identity"
+}
```
lorenyu, line 9: `data` modules should only have a single variable to make them simple. The whole point is that using the data module should be easier than using the data sources directly. Per my previous comment about the `resources` module let's get rid of `name` and keep `domain_name`.

infra/modules/notifications/resources/main.tf
```diff
@@ -0,0 +1,77 @@
...
+locals {
+  mail_from_domain          = "mail.${var.domain_name}"
+  stripped_domain_name      = replace(var.domain_name, "/[.]$/", "")
+  stripped_mail_from_domain = replace(local.mail_from_domain, "/[.]$/", "")
+  dash_domain               = replace(var.domain_name, ".", "-")
+}
+
+# Verify email sender identity.
+# Docs: https://docs.aws.amazon.com/pinpoint/latest/userguide/channels-email-manage-verify.html
+resource "aws_sesv2_email_identity" "sender" {
```
lorenyu, line 14: Per my comment about renaming this module to `notifications-email-domain`, let's rename this resource from "sender" to "sender_domain"

infra/modules/notifications/resources/outputs.tf
```diff
@@ -0,0 +1,7 @@
...
+
+output "email_identity_arn" {
```
lorenyu, line 5: Let's rename this to "domain_identity_arn"

lorenyu (commented):
Ok I think I'm officially done with my review. Can take another pass once we address the changes. Thanks for your patience!

infra/app/service/notifications.tf
```diff
@@ -0,0 +1,50 @@
...
+    !local.is_temporary ? module.notifications[0].email_identity_config : module.existing_notifications[0].email_identity_config
+  ) : null
+  notifications_environment_variables = module.app_config.enable_notifications ? {
+    PINPOINT_APP_ID = module.notifications_app[0].app_id,
+    PINPOINT_SENDER = local.notifications_config.sender_email
```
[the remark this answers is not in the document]
coilysiren: Easy to change it now! I'm do that

infra/modules/notifications/data/main.tf
coilysiren: Huh! Not sure why my source examples had it then. Good catch

infra/modules/notifications/resources/variables.tf
coilysiren: unfortunately not possible:

```
│ Error: Invalid reference in variable validation
│
│   on ../../modules/notifications/resources/variables.tf line 16, in variable "sender_email":
│   16:     condition     = can(regex("^.+@${var.domain_name}$", var.sender_email))
│
│ The condition for variable "sender_email" can only refer to the variable itself, using var.sender_email.
```

infra/modules/notifications/resources/main.tf
```diff
@@ -0,0 +1,77 @@
...
+    reputation_metrics_enabled = true
+  }
+
+  sending_options {
+    sending_enabled = true
+  }
+
+  suppression_options {
+    suppressed_reasons = ["BOUNCE", "COMPLAINT"]
+  }
```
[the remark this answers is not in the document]
coilysiren: I honestly don't know why this wasn't using the default values! I removed it

infra/modules/notifications/resources/main.tf
coilysiren: I'll see if that works!

infra/modules/notifications/resources/variables.tf
lorenyu: Looks like it's possible in Terraform 1.9.0
https://github.com/hashicorp/terraform/blob/v1.9/CHANGELOG.md#190-june-26-2024

I'm fine upgrading Terraform version now or punting this to follow up PR if you want. Defer to you. If you punt to a follow up PR can you create a ticket?

coilysiren: Yeah I'll make a ticket for this one

coilysiren:
It took me an undue amount of time to get the `name => domain_name` thing right 😓

infra/modules/notifications/resources/main.tf
coilysiren: It works!

infra/app/service/identity_provider.tf
coilysiren: Made a ticket for this one https://github.com/navapbc/template-infra/issues/778

infra/modules/notifications/resources/variables.tf
coilysiren: ticket: https://github.com/navapbc/template-infra/issues/777

infra/modules/notifications-app/resources/variables.tf
coilysiren: I mean, we can. But I think that would cause confusion when looking at the module inputs. You would see `sender_display_name` but not `sender_email`, and possibly wonder why you are inputting a display name if you aren't also inputting an email.

coilysiren: I think it could make sense if we made display name more generically named so it looks like this:

```hcl
module "notifications" {
  count  = module.app_config.enable_notifications ? 1 : 0
  source = "../../modules/notifications/resources"

  name                = local.notifications_app_name
  domain_identity_arn = local.domain_identity_arn
  display_name        = local.notifications_config.display_name
}
```

Then you would use display name in other places... I think.

coilysiren: I wonder if we should make display name a required attribute... it would make this line much simpler

```hcl
  from_address   = var.sender_email != null ? (var.display_name != null ? "${var.display_name} <${var.sender_email}>" : var.sender_email) : null
```

coilysiren: Hmm, I actually don't see this

> a data source with the email_identity_arn and doing data.email_identity_arn.sender.email

here?

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/sesv2_email_identity

coilysiren: Ah wait, I actually don't see this

> a data source with the email_identity_arn and doing data.email_identity_arn.sender.email

here?

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/sesv2_email_identity

Unless there's some terraform knowledge that I'm missing

lorenyu (commented):
code LGTM but couple of comments:
1. can you test the identity provider flows to make sure the account creation emails are still sending properly since we made changes to identity provider
2. i think i spotted a bug in the sesv2_email_identity resource where we are passing in the dash_domain for the email_identity

potentially out of scope for this PR, but the identity provider and notifications features are sufficiently complex that i think we should add some e2e testing (in template-only-ci-infra.yml) for those to prevent regressions, maybe we can do that in a later PR and just be ok with manual testing for now

infra/modules/notifications-email-domain/resources/main.tf
```diff
@@ -0,0 +1,40 @@
...
+locals {
+  mail_from_domain          = "mail.${var.domain_name}"
+  stripped_mail_from_domain = replace(local.mail_from_domain, "/[.]$/", "")
+  dash_domain               = replace(var.domain_name, ".", "-")
+}
+
+# Verify email sender identity.
+# Docs: https://docs.aws.amazon.com/pinpoint/latest/userguide/channels-email-manage-verify.html
+resource "aws_sesv2_email_identity" "sender_domain" {
+  email_identity         = local.dash_domain
```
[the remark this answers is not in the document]
coilysiren: Yep, it was incorrect https://github.com/navapbc/platform-test/pull/140/commits/e82004499fb579878d64f6931c81572444701042

infra/modules/notifications/resources/email.tf
```diff
@@ -0,0 +1,5 @@
+resource "aws_pinpoint_email_channel" "app" {
+  application_id = aws_pinpoint_app.app.application_id
+  from_address   = var.sender_email != null ? (var.sender_display_name != null ? "${var.sender_display_name} <${var.sender_email}>" : var.sender_email) : null
```
lorenyu, line 3: if i understand correctly sender_email is a required variable, so can we simplify this to a single conditional like:

```suggestion
  from_address   = var.sender_display_name != null ? "${var.sender_display_name} <${var.sender_email}>" : var.sender_email)
```

infra/modules/notifications-email-domain/resources/variables.tf
```diff
@@ -0,0 +1,15 @@
...
+variable "sender_email" {
+  type        = string
+  description = "Email address to use to send notification emails"
+}
+
+variable "sender_display_name" {
+  type        = string
+  description = "The display name for notification emails. Only used if sender_email is provided"
+  default     = null
+}
```
lorenyu, line 15: Remove these unused variables

infra/app/service/notifications.tf
```diff
@@ -0,0 +1,46 @@
...
+
+# If the app has `enable_notifications` set to true AND this is not a temporary
+# environment, then create a email notification identity.
+module "notifications_email_domain" {
+  count  = module.app_config.enable_notifications && !local.is_temporary ? 1 : 0
+  source = "../../modules/notifications-email-domain/resources"
+
+  domain_name         = local.service_config.domain_name
+  sender_email        = local.notifications_config.sender_email
+  sender_display_name = local.notifications_config.sender_display_name
```
lorenyu, line 23: I don't think these variables are needed

lorenyu (approved):
I think this is ready. Just some nits

infra/app/service/notifications.tf
coilysiren: Wow... these module signatures are looking clean as heck

lorenyu: :) haha that's basically the thing I care about the most when it comes to the platform design. It's easier to add things to interfaces than change or remove them so I like to keep them as minimal as possible.

coilysiren:
test:

image

lorenyu:
🚀

PR #143, opened 2024-12-13
Add workflow that scans orphaned PR environments

lorenyu:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/709

## Changes

- Add bin/orphaned-prs
- Add scan-orphaned-pr-environments.yml workflow
- Pulled out get_app_names into reusable function in util.bash

## Context for reviewers

I also pulled out get_app_names function from infra-deploy-status-check-configs into a utils.bash library file.

### testing orphaned PR scan
As luck may have it we already have an orphaned PR environment so I was able to do a test showing orphaned PR environments: https://github.com/navapbc/platform-test/actions/runs/12601493551

which [shows up on Slack like this](https://nava.slack.com/archives/C03G1SWD9H7/p1734656061200609)
image

### regression testing infra deploy status check
I also ran infra deploy status check to make sure there are no regressions: https://github.com/navapbc/platform-test/actions/runs/12601495350

### testing when systems notification config is empty
the "Notify" job still runs fine, it just ends up being a no-op https://github.com/navapbc/platform-test/actions/runs/12590423737/job/35091916665

- Service endpoint: http://p-143-app-dev-1581364616.us-east-1.elb.amazonaws.com
- Deployed commit: 09f9a512f9d5c77ece6f32d9b49dbaa61671ebf3


bin/infra-deploy-status-check-configs
```diff
@@ -47,6 +47,8 @@
 # -----------------------------------------------------------------------------
 set -euo pipefail
 
+source bin/util.bash
```
coilysiren, line 50: This one ends in `.bash` but the other ones don't?

bin/orphaned-prs
```diff
@@ -0,0 +1,45 @@
...
+# resources in the workspace should have been destroyed and the workspace
+# deleted, so existing workspaces for closed PRs are considered orphaned.
+# -----------------------------------------------------------------------------
+set -euo pipefail
+
+app_name="$1"
+
+echo "::group::Initialize Terraform"
+echo terraform -chdir="infra/${app_name}/service" init -input=false -reconfigure -backend-config="dev.s3.tfbackend"
+terraform -chdir="infra/${app_name}/service" init -input=false -reconfigure -backend-config="dev.s3.tfbackend"
```
coilysiren, line 14: I know this is probably a convention used in dozens of places, but why not use `set -xtrace` here?

bin/orphaned-prs
```diff
@@ -0,0 +1,45 @@
...
+echo "::endgroup::"
+
+echo "::group::List PRs with PR environments"
```
coilysiren, line 17:
```suggestion
echo "::group::PR environments, which corresponds to a list of PRs"
```

.github/workflows/send-system-notification.yml
```diff
@@ -51,6 +54,10 @@ jobs:
             slack_token_secret_name="$(echo "${channel_config}" | jq -r ".slack_token_secret_name")"
             echo "Slack token secret name: ${slack_token_secret_name}"
             echo "SLACK_TOKEN_SECRET_NAME=${slack_token_secret_name}" >> "$GITHUB_ENV"
+
+            # Convert Markdown links in message [text](url) to Slack format <url|text>
+            echo "Convert message from Markdown to Slack format"
+            echo "SLACK_MESSAGE=$(echo "${{ inputs.message }}" | sed -E 's/\[(.+)\]\((.+)\)/<\2|\1>/g')" >> "$GITHUB_ENV"
```
coilysiren, line 60: I would expect to see this pipe sed bit down lower, on the `payload: | ... text: ` line. I assume there was some constraint that make you want to pipe it into an environment variable instead?

coilysiren: Oh, its that the content of the `payload` value aren't bash. I'm just gonna leave this comment here in case it illuminates anyone else.

coilysiren (approved):
I have small feedbacks

bin/infra-deploy-status-check-configs
lorenyu: following google's shell style guide https://google.github.io/styleguide/shellguide.html#file-extensions

> Executables should have no extension (strongly preferred) or a .sh extension. Libraries must have a .sh extension and should not be executable.
>
> It is not necessary to know what language a program is written in when executing it and shell doesn’t require an extension so we prefer not to use one for executables.
>
> However, for libraries it’s important to know what language it is and sometimes there’s a need to have similar libraries in different languages. This allows library files with identical purposes but different languages to be identically named except for the language-specific suffix.

bin/orphaned-prs
lorenyu: is it possible to do that for just one command? i wouldn't want to echo the `echo "::group::..."` lines

bin/orphaned-prs
lorenyu: The suggestion doesn't seem accurate. What we are listing are PR numbers for PRs that have an active PR environment.

bin/orphaned-prs
doshitan, line 1: I'd suggest a slightly more precise name, like `bin/orphaned-pr-environments`. From the name alone, I might expect this is checking if the PR author's account is still in the org/active, if there's been activity on the PR in the last X months, etc.

bin/infra-deploy-status-check-configs
doshitan: nit, though following those guidelines it should be `bin/util.sh` not `bin/util.bash` right?

> Libraries must have a .sh extension

bin/orphaned-prs
```diff
@@ -0,0 +1,45 @@
...
+  if [ "$pr_status" == "CLOSED" ]; then
+    closed_prs+=("$pr_num")
+  fi
+done
+echo "::endgroup::"
+
+# if closed_prs is not empty exit with 1 otherwise exit with 0
+if [ ${#closed_prs[@]} -gt 0 ]; then
+  echo "Orphaned PR environments: ${closed_prs[*]}"
+  echo "Orphaned PR environments: ${closed_prs[*]}" >> "${GITHUB_STEP_SUMMARY}"
```
doshitan, line 40: I know we are in the context of PRs, with IDs listed in previous steps, but I would suggest clarifying these are PR IDs somehow, because looking at the output like:

```
Orphaned PR environments: 140
```

Reads to me that there are 140 orphaned environments! Would probably be clearer when there are multiple IDs together, but the bare number might still be confusing.

Maybe:
```suggestion
  echo "Orphaned PR environments for the following PRs: ${closed_prs[*]}" >> "${GITHUB_STEP_SUMMARY}"
```

doshitan (approved):

lorenyu:
@doshitan not sure why i couldn't comment inline on this one so creating a new comment

> nit, though following those guidelines it should be bin/util.sh not bin/util.bash right?
>
> > Libraries must have a .sh extension

I was a little confused by the guidance, I saw this follow up paragraph which seemed to imply that libraries should have a language specific extension:

> However, for libraries it’s important to know what language it is and sometimes there’s a need to have similar libraries in different languages. This allows library files with identical purposes but different languages to be identically named except for the language-specific suffix. But maybe I'm misinterpreting?

lorenyu:
Done in https://github.com/navapbc/template-infra/pull/819

doshitan:
> I was a little confused by the guidance, I saw this follow up paragraph which seemed to imply that libraries should have a language specific extension:
> 
> > However, for libraries it’s important to know what language it is and sometimes there’s a need to have similar libraries in different languages. This allows library files with identical purposes but different languages to be identically named except for the language-specific suffix. But maybe I'm misinterpreting?

My read is they mean to differentiate `lib.sh` from `lib.py`/`lib.rs`/`lib.js`, not `lib.zsh`/`lib.bash`/`lib.fish` (though there might be times when it could make sense to do that as well).

lorenyu:
> > I was a little confused by the guidance, I saw this follow up paragraph which seemed to imply that libraries should have a language specific extension:
> > > However, for libraries it’s important to know what language it is and sometimes there’s a need to have similar libraries in different languages. This allows library files with identical purposes but different languages to be identically named except for the language-specific suffix. But maybe I'm misinterpreting?
> 
> My read is they mean to differentiate `lib.sh` from `lib.py`/`lib.rs`/`lib.js`, not `lib.zsh`/`lib.bash`/`lib.fish` (though there might be times when it could make sense to do that as well).

Got it, I renamed to .sh

PR #146, opened 2024-12-24
Fix race condition when publishing images

doshitan:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/836

## Changes

The `ci-{{app_name}}-infra-service.yml` and `cd-{{app_name}}.yml` workflows
occasionally run into issues trying to publish an already existing image tag[1],
failing with something like:

> tag invalid: The image tag '48995925face06c5fcb68fc52c50e7d57bb46ff6' already exists in the 'platform-nextjs-app' repository and cannot be overwritten because the repository is immutable.

I believe what's happening is that for merges to `main`, the two workflows get
kicked off at the same time, both needing build and publish an image for the
same commit and using different methods for doing that building/publishing.
`ci-{{app_name}}-infra-service.yml` runs the commands internally to its test
logic, `cd-{{app_name}}.yml` uses the `build-and-publish.yml` workflow.

By having `ci-{{app_name}}-infra-service.yml` use the re-usable
build-and-publish workflow that `cd-{{app_name}}.yml` is using, which has an
internal concurrency limit, the workflows should no longer race against each
other to build the same commit.

[1] https://github.com/navapbc/platform-test-nextjs/actions/runs/12484891877/job/34842946860

## Testing

Not sure how to reliably reproduce the race condition. Could manually dispatch the workflows at "the same time" (as close as is physically possible to click two different buttons in different windows at the same time)?

If it works, any other workflows that should be using the build-and-publish workflow as well?

- Service endpoint: http://p-146-app-dev-1640477009.us-east-1.elb.amazonaws.com
- Deployed commit: 35415ce04664d94d63e68dbe4c60cdd0e6e364e4

- Service endpoint: http://p-146-app-rails-dev-1056090106.us-east-1.elb.amazonaws.com
- Deployed commit: 35415ce04664d94d63e68dbe4c60cdd0e6e364e4


.github/workflows/ci-app-infra-service.yml
```diff
@@ -16,11 +16,26 @@ on:
...
 
 jobs:
+  # `make infra-test-service` will build and publish the release itself, but
+  #  doing it explicitly with the re-usable workflow helps avoid race conditions
+  build-and-publish:
+    name: Build
+    uses: ./.github/workflows/build-and-publish.yml
+    with:
+      app_name: app
+      ref: ${{ inputs.version || github.event.pull_request.head.sha || 'main' }}
```
lorenyu, line 33: nit: `github.event.pull_request.head.sha` only works for the `pull_request` trigger and won't work for the `push` trigger. It's coincidence that the hardcoded `'main'` ref works for the `push` trigger since the `push` trigger coincidentally is configured to only trigger with `branches: [main]`, but that's not as DRY as it could be. Consider adding `github.ref` to this chain, potentially just replacing `github.event.pull_request.head.sha`.

```suggestion
      ref: ${{ inputs.version || github.ref || 'main' }}
```

or 

```
      ref: ${{ inputs.version || github.event.pull_request.head.sha || github.ref || 'main' }}
```

.github/workflows/ci-app-infra-service.yml
lorenyu, line 1: since we added the ability to run this workflow with a given `version`, we need to pass that version to [actions/checkout](https://github.com/actions/checkout) since actions/checkout by default checks out the ref that triggered the workflow or main, which may not match the version being tested.

.github/workflows/ci-app-infra-service.yml
```diff
@@ -16,11 +16,26 @@ on:
...
   workflow_dispatch:
+    inputs:
+      version:
+        required: true
+        default: "main"
+        description: Tag or branch or SHA to test
 
 jobs:
+  # `make infra-test-service` will build and publish the release itself, but
+  #  doing it explicitly with the re-usable workflow helps avoid race conditions
```
lorenyu, line 27: Actually thinking about this more, seems like with this change we should remove the build and publish code from infra-test-service and have that infra-test-service rely on the fact that it's already built. Infra test service is meant to just test the terraform apply and terraform destroy anyways. The build and publish can be done outside of that test suite.

Benefit of removing is to avoid making this workflow take longer by having to build twice

lorenyu:
created a ticket for this for myself to track https://github.com/navapbc/template-infra/issues/836

lorenyu:
Just making a note that this race condition seems to happen more commonly than I originally realized.

lorenyu:
Done in https://github.com/navapbc/template-infra/pull/848

PR #150, opened 2025-01-07
Send failure notifications on CI/CD workflows

lorenyu:
## Ticket

## Changes

## Context for reviewers

## Testing

- Service endpoint: http://p-150-app-dev-778906576.us-east-1.elb.amazonaws.com
- Deployed commit: eba7891efd0b5cb0105e6bbe6753095f5f866f1d


lorenyu:
At first I thought it would be useful since I was under the impression that when pull requests are merged to main and those merges cause a CI failure that the person that would get notified was the last person who touched the workflow, but I think that's not true. I think the person that gets notified is the person that merged the PR. In the case of platform-test* changes, that would typically be platform-bot, who's email is set to platform-admins@navapbc.com, so we should get the email notifications.

PR #153, opened 2025-01-14
Move prefix from env-config to service root

lorenyu:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/654

## Changes

- Move prefix local var from env-config module to service layer's root module
- Move storage resources to separate storage.tf file

## Context for reviewers

See ticket

## Testing

1. Running `make infra-update-app-service APP_NAME=app ENVIRONMENT=dev` shows no changes:

   image

2. Inspecting PR environment that was created shows that service and storage bucket still have prefix

   image

   image

- Service endpoint: http://p-153-app-dev-86780802.us-east-1.elb.amazonaws.com
- Deployed commit: c34d29ca8ac081ee4cb86dc664ef283e5f1ce804

### DIFF
--- a/bin/update-pr-environment
+++ b/bin/update-pr-environment
@@ -18,6 +18,7 @@ pr_number="$3"
 image_tag="$4"
 
 workspace="p-${pr_number}"
+prefix="${workspace}-"
 
 echo "::group::Initialize Terraform with backend for environment: ${environment}"
 terraform -chdir="infra/${app_name}/service" init -backend-config="${environment}.s3.tfbackend"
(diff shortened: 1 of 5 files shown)


coilysiren (approved):
I have a small worry that this might result in resource name issues (eg. things suddenly have a perfix in them, or vice versa). But if that happens it'll almost certainly happen inside of PR environments. 

At any rate, the code looks good

lorenyu:
> I have a small worry that this might result in resource name issues (eg. things suddenly have a perfix in them, or vice versa). But if that happens it'll almost certainly happen inside of PR environments.
> 
> At any rate, the code looks good

the terraform plan shows no changes, and the PR environment in this PR also shows that the temp environment has the correct name for bucket and service still

i think the bin scripts don't have coverage from this, but i looked up service_name and saw that the only other scripts that use service_name are the deploy script and run-command. deploy script i think only makes sense to use for default workspace. and run-command we are gonna add ability to run on PR environment but that's in a ticket, so for now i think it's ok that both of those assume default workspace.

lorenyu:
done in https://github.com/navapbc/template-infra/pull/832

PR #154, opened 2025-01-16
Fix notifications access from service

lorenyu:
## Ticket

Contributes to https://github.com/navapbc/template-infra/issues/792

## Changes

Fix ability for ECS service to access Pinpoint:
- Add VPC endpoints for Pinpoint services
- Add IAM policy for accessing Pinpoint and SES to service

Also add notifications test endpoint to example app for testing:
- Add /notifications page where you can enter an email address and get a test notification to that email address
- Add links to the homepage to all the test pages in the example app

## Context for reviewers

I added /notifications endpoint since it will be useful to test changes to notifications infrastructure, but while testing I discovered some access bugs

## Testing

Applied changes to dev network to create the two VPC endpoints using
```
make infra-update-network NETWORK_NAME=dev
```

Tested locally
image

Tested on the PR environment, you can too! (took a while to get the IAM permissions right)
image

- Service endpoint: http://p-154-app-dev-2103568609.us-east-1.elb.amazonaws.com
- Deployed commit: f6d6a07a6ddec0a9c25682a1247d4ea1576d8cf8

### DIFF
--- a/infra/modules/network/vpc_endpoints.tf
+++ b/infra/modules/network/vpc_endpoints.tf
@@ -18,6 +18,9 @@ locals {
 
     # AWS services used by ECS Exec
     var.enable_command_execution ? ["ssmmessages"] : [],
+
+    # AWS services used by notifications
+    var.enable_notifications ? ["pinpoint", "email-smtp"] : [],
   )
 
   # S3 and DynamoDB use Gateway VPC endpoints. All other services use Interface VPC endpoints
--- a/infra/modules/notifications/resources/access_control.tf
+++ b/infra/modules/notifications/resources/access_control.tf
@@ -1,6 +1,6 @@
 resource "aws_iam_policy" "access" {
-  name        = "pinpoint_policy"
-  description = "Policy for calling SendMessages and SendUsersMessages on Pinpoint app"
+  name        = "${var.name}-access"
+  description = "Policy for calling SendMessages and SendUsersMessages on Pinpoint app ${var.name}"
 
   policy = jsonencode({
     Version = "2012-10-17"
--- a/infra/modules/notifications/resources/access_control.tf
+++ b/infra/modules/notifications/resources/access_control.tf
@@ -12,8 +12,7 @@ resource "aws_iam_policy" "access" {
           "mobiletargeting:SendMessages",
           "mobiletargeting:SendUsersMessages"
         ]
-        Resource = "*"
-        # Resource = "${aws_pinpoint_app.app.arn}/messages"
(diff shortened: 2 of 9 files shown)


infra/modules/network/vpc_endpoints.tf
```diff
@@ -18,6 +18,9 @@ locals {
 
     # AWS services used by ECS Exec
     var.enable_command_execution ? ["ssmmessages"] : [],
+
+    # AWS services used by notifications
+    var.enable_notifications ? ["pinpoint", "email-smtp"] : [],
```
coilysiren, line 23: It's called `mobile-targeting` [in IAM](https://github.com/navapbc/template-infra/blob/cb15833ca914af837d380ceec3db9aae432dc811/infra/project-config/aws_services.tf#L21) but the VPC endpoints are `["pinpoint", "email-smtp"]` ??? bah

lorenyu: yeah i put mobiletargeting at first and got an error saying that doesn't exist.

also, the AWS docs don't mention "email-smtp" anywhere, instead they said to search for smtp and then click on that one, so I had to go to the console and search to find out that email-smtp was what they meant.

infra/modules/notifications/resources/access_control.tf
```diff
@@ -0,0 +1,32 @@
...
+      # From https://docs.aws.amazon.com/pinpoint/latest/developerguide/permissions-ses.html
+      {
+        Effect = "Allow"
+        Action = [
+          "ses:SendEmail",
+          "ses:SendRawEmail",
+        ]
+        Resource = [
+          var.domain_identity_arn,
+          "arn:*:ses:*:*:configuration-set/*",
```
coilysiren, line 27: Any specific reason why this 2nd on isn't an arn as well?

infra/modules/network/vpc_endpoints.tf
lorenyu: image
https://docs.aws.amazon.com/ses/latest/dg/send-email-set-up-vpc-endpoints.html

amazing right?

coilysiren (approved):
non-blocking q

infra/modules/notifications/resources/access_control.tf
coilysiren: Wow... yeah no arns mentioned [here](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonses.html). That's a new one for me!

lorenyu:
done in https://github.com/navapbc/template-infra/pull/835

PR #161, opened 2025-01-21
Nest network and within service & database modules

coilysiren:
## Motivation and Context

The motivation of this PR is in this diff:

image

Nesting the network module within the service/database modules has the following impact:

- reduces the networking-related API surface of those modules
- abstract the remaining networking API surface away from AWS

This makes the modules more flexible and resilient!

## Testing

```
~/projects/platform-test/infra/app/service/

$ terraform plan -var environment_name=dev

> No changes. Your infrastructure matches the configuration.
```

```
~/projects/platform-test/infra/app/database/

$ terraform plan -var environment_name=dev

> No changes. Your infrastructure matches the configuration.
```

- Service endpoint: http://p-161-app-dev-1734989274.us-east-1.elb.amazonaws.com
- Deployed commit: 429790fb6df97cbd4c8dbe34214a3a111ee3838c

- Service endpoint: http://p-161-app-rails-dev-1481459668.us-east-1.elb.amazonaws.com
- Deployed commit: 429790fb6df97cbd4c8dbe34214a3a111ee3838c


infra/modules/network/data/main.tf
```diff
@@ -3,30 +3,34 @@ module "interface" {
   name   = var.name
 }
 
+module "project_config" {
```
lorenyu, line 6: as we discussed in our 1:1 don't introduce dependencies on config modules in reusable modules

lorenyu (commented):
nice!

infra/modules/network/data/main.tf
coilysiren: I can definitely make that change! I just totally totally remember the motivation there. Like, that's the conceptual difference between nesting the network model and nesting the project config module?

lorenyu: infra/modules/ are reusable modules. they should have no dependencies on anything project specific.
Longer term once these modules are stable we may decided to [publish them](https://developer.hashicorp.com/terraform/registry/modules/publish) to the terraform registry, so we should have no dependencies on any local project configuration.

coilysiren: Ah! So infra/modules are truly meant to be as flexible as terraform modules you would find on the registry. Gotcha. 

That changes my point of view here a bit... but I'm going to continue on with nesting the VPC module

lorenyu: Yeah that's also why I was on the fence about nesting the network module, because by nesting the network module it means that database and service modules now have a dependency on the network module, meaning that in order for someone to use the database and service modules they would also need to use the network module, or at least adhere to the same interface for the network module. But after seeing how it cleans up the interface I think maybe the benefit is worth the tradeoff.

infra/app/service/main.tf
```diff
@@ -63,10 +63,7 @@ module "service" {
 
   image_tag = local.image_tag
 
-  vpc_id                         = module.network.vpc_id
-  public_subnet_ids              = module.network.public_subnet_ids
-  private_subnet_ids             = module.network.private_subnet_ids
-  aws_services_security_group_id = module.network.aws_services_security_group_id
```
coilysiren, line 69: 😎

infra/modules/service/main.tf
```diff
@@ -33,6 +33,10 @@ locals {
   )
 }
 
+module "project_config" {
```
lorenyu, line 36: clean this up

lorenyu (approved):

lorenyu (approved):

lorenyu:
Done in https://github.com/navapbc/template-infra/commit/eddae5cccae64eb26f18a045b6884bf84512ec84

PR #169, opened 2025-01-31
New image scanning workflow and parsing

Nava-JoshLong:
## Ticket

## Changes

- Created python scripts to parse JSON files from scanners, and create markdown for easier viewing of results
- Created new workflows for the new image scanning logic
- Create action for image scan

## Context for reviewers

I put things in parallel to the existing workflows so I can test if these changes work, without messing with any subsequent runs. I had to tweak the files from the contract's version since it has things more specific to our setup and aren't needed here.

The main changes here will make all image scans save findings to JSON, and then a python script to parse and format it into a clean markdown file that we can view in the workflow summary, and use for creating Github tickets. The ticket creation and Slack alerting is disabled by default, with comments on what it takes to enable. 

There's some side benefits of this, such as handling if the image is ARM based, and better trivy finding ignore functionality. There is a change to the current method of parallel scanning, to being a single build and use that image on all of the scanners. This is because as images grow, the time it takes to build also increase, and it doesn't make a lot of sense to build it multiple times (IMHO). In real life, the CI unit and coverage tests takes the same time, or longer than this change

## Testing

Testing will be handled by the PR running these workflows

- Service endpoint: http://p-169-app-dev-1653289684.us-east-1.elb.amazonaws.com
- Deployed commit: d8f4fbe8c5a9ee984581f04a58bef7ea99ac6559

- Service endpoint: http://p-169-app-rails-dev-663016209.us-east-1.elb.amazonaws.com
- Deployed commit: d8f4fbe8c5a9ee984581f04a58bef7ea99ac6559

### DIFF
--- a/.github/workflows/ci-vulnerability-scans.yml
+++ b/.github/workflows/ci-vulnerability-scans.yml
@@ -1,0 +1,82 @@
+# This workflow will scan the release versions of your Docker images, and standardize the results into a JSON file. This file
+# can be used to create Github tickets and Slack alerts
+
+name: Template Release Image Scans
+
+on:
+  schedule:
+    - cron: '0 0 * * *'  # Run daily at midnight UTC
+  pull_request:
+    paths:
+      # Handles when there are changes to the ignore files
+      - .grype.yml        # Anchore ignore
+      - .hadolint.yaml    # Hadolint ignore
+      - .trivyignore.yaml # Trivy ignore
+      - trivy.yaml        # Trivy config that points to the ignore file
+      # Handles when there are changes to the actual workflows we use
+      - .github/workflows/ci-vulnerability-scans.yml
+      - .github/workflows/vulnerability-scans.yml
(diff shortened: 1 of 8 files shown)


.github/workflows/ci-vulnerability-scans.yml
```diff
@@ -0,0 +1,82 @@
...
+# permissions:
+#     contents: read
+#     issues: write
+
+jobs:
+  matrix_image_scans:
+    # Setting to new in case there's anything checking the existing vulnerability-scans.yml
+    # This will be fixed before merging
+    uses: ./.github/workflows/new-vulnerability-scans.yml
+    strategy:
```
Nava-JoshLong, line 44: This might be the issue from [here](https://github.com/navapbc/platform-test/actions/runs/13081157307). In my file I pulled this from, I have the matrix in the workflow this calls, but I forgot why I did that. I think this could be why I did it like that. If so, should be an easy move into the sub-workflow to test it

PR #172, opened 2025-02-05
enable local login with devise

### LINKED ISSUE #171: Enable local login via devise
To unblock local development when developer lacks AWS credentials, it would be helpful to have the ability to log in locally

LeslieJAnderson:
## Ticket

Resolves #171

## Changes

Adds the ability to create local users and log in locally via devise

## Context for reviewers

To unblock local development when developer lacks AWS credentials, it would be helpful to have the ability to log in locally.

## Testing

Add `USE_DEVISE=true` to your `.env` file

To test: Run the application, then create a user via the UI. It will take you to the screen that claims you will be receiving an email. You will not receive that email, however, if you go back to the main page and attempt to log in using those credentials, it should work.

- Service endpoint: https://p-172-app-dev-1902816801.us-east-1.elb.amazonaws.com
- Deployed commit: 0f37c436faeaf47196ec8cf4975a85ab3c84f3a6

- Service endpoint: https://p-172-app-rails-dev-185120163.us-east-1.elb.amazonaws.com
- Deployed commit: 0f37c436faeaf47196ec8cf4975a85ab3c84f3a6


app-rails/local.env.example
```diff
@@ -33,6 +33,7 @@ AWS_DEFAULT_REGION=<FILL ME IN>
 COGNITO_USER_POOL_ID=<FILL ME IN>
 COGNITO_CLIENT_ID=<FILL ME IN>
 COGNITO_CLIENT_SECRET=<FILL ME IN>
+USE_DEVISE=true
```
CristianAThompson, line 36: We should also include the 

`AUTH_ADAPTER=mock`

Env variable here in the local to ensure by default it's spinning up with the MockAdapter from the Factory

LeslieJAnderson: Good catch, thank you!

.template-infra/app-app-rails.yml
```diff
@@ -4,4 +4,4 @@ _src_path: https://github.com/navapbc/template-infra
 app_has_dev_env_setup: true
 app_local_port: 3100
 app_name: app-rails
-template: app
+template: app
```
lorenyu, line 7: i'm confused why this shows up as a diff, looks the same

app-rails/README.md
```diff
@@ -96,6 +96,29 @@ To run natively:
...
 1. Then visit http://localhost:3100
 
+### Enabling Local Login for Development
+
+For local development (without AWS Cognito credentials), you can use local login via Devise. To enable local login:
+
+1. **Set Rails environment to development**:
+
+   ```bash
+   RAILS_ENV=development
```
lorenyu, line 106: Isn't this already set by default?

app-rails/README.md
```diff
@@ -96,6 +96,29 @@ To run natively:
...
+
+2. **Enable Devise in development by setting the environment variable:**:
+   ```bash
+   USE_DEVISE=true
```
lorenyu, line 110: Is this needed? When would we ever _not_ want to use devise?

app-rails/README.md
```diff
@@ -96,6 +96,29 @@ To run natively:
...
+
+3. **Set up the database (run migrations):**:
+   ```bash
+   rails db:migrate
+
+4. **Follow the instructions for starting a container then start it:**:
+   ```bash
+   make start-container
```
lorenyu, line 118: these two instructions seem redundant with instructions for just running the app locally

app-rails/db/migrate/20250414194640_add_local_devise_fields_to_users.rb
lorenyu, line 1: The issue with this approach is this migration will also run in other environments including production. It will be confusing to have columns that are not used in any environment except for local development, especially with columns as seemingly important as encrypted passwords

lorenyu (commented):
At a high level, this approach is a full implementation of authentication, using a database to store authentication credentials. It's too complex for what the goal of this ticket is and also has the added problem of creating unused database fields in production environments. A potential approach that might be better would be to implement a MockAuthAdapter that has a hardcoded password or an otherwise simple authentication challenge that is easy to pass (e.g. it always returns true) rather than a full and real authentication implementation.

PR #173, opened 2025-02-05
Change references to ="$(expression)" from =$(expression) in scripts

daphnegold:
## Ticket

Resolves #669
Resolves #679

## Changes

- Shellcheck: I investigated this, and added some optional rules to our shellcheck command in the makefile. It's not enforced consistently and just a warning, as shellcheck works dynamically making decisions. I don't think it's a stringent enough linter for our desires to enforce double-quotes in our code style, unfortunately.
- Add double quotes around `$()` and `${}`

Note: We should consider whether we want to standardize double quotes around `${{ }}` in GitHub Actions. I don’t think it affects word splitting or globbing, but we are currently inconsistent with it.

- Service endpoint: http://p-173-app-dev-476427417.us-east-1.elb.amazonaws.com
- Deployed commit: d378abc6348b9fc0ee16a89202c597e878c2d802

- Service endpoint: http://p-173-app-rails-dev-1202887749.us-east-1.elb.amazonaws.com
- Deployed commit: d378abc6348b9fc0ee16a89202c597e878c2d802


.github/actions/first-file/action.yml
```diff
@@ -4,7 +4,6 @@ description: Check given list of files in order and return first one that exists
 inputs:
   files:
     required: true
-    type: string
```
daphnegold, line 7: This was causing an alert that `type: string` isn't allowed here

PR #175, opened 2025-02-07
Support ephemeral write mounts for services

doshitan:
## Ticket

Related to https://github.com/navapbc/template-application-rails/issues/58

## Changes

Rather than support/encourage a writable root filesystem, support explicitly defining locations that are writable, which is also helpful for development purposes.

## Context for reviewers

- `app-rails/` changes go to `template-application-rails`
- `infra/modules/` changes go to `template-infra`
- Common bits of `infra/app/` and `infra/app-rails/` go to `template-infra`
- `infra/app-rails/` customizations apply to this repo (after previous changes are merged to respective repos, then merged to this one)

## Testing

See CI.

Clicked around PR Env.

- Service endpoint: http://p-175-app-dev-808286377.us-east-1.elb.amazonaws.com
- Deployed commit: 9410bfd3d9d4cd8d92e4eb5621e1d389e32a0a81

- Service endpoint: http://p-175-app-rails-dev-1338933003.us-east-1.elb.amazonaws.com
- Deployed commit: 9410bfd3d9d4cd8d92e4eb5621e1d389e32a0a81

### DIFF
--- a/infra/app-rails/service/main.tf
+++ b/infra/app-rails/service/main.tf
@@ -125,9 +125,7 @@ module "service" {
     } : {},
   )
 
-  ephemeral_write_volumes = [
-    "/rails/tmp"
-  ]
+  ephemeral_write_volumes = local.service_config.ephemeral_write_volumes
 
   is_temporary = local.is_temporary
 
--- a/infra/app-rails/service/main.tf
+++ b/infra/app-rails/service/main.tf
@@ -125,6 +125,10 @@ module "service" {
     } : {},
   )
 
+  ephemeral_write_volumes = [
+    "/rails/tmp"
+  ]
+
   is_temporary = local.is_temporary
 
   # Template Divergent Variables
--- a/infra/app-rails/service/main.tf
+++ b/infra/app-rails/service/main.tf
@@ -128,6 +128,6 @@ module "service" {
   is_temporary = local.is_temporary
 
   # Template Divergent Variables
-  container_read_only = false
+  container_read_only = true
   healthcheck_type    = "curl"
 }
--- a/infra/app-rails/service/main.tf
+++ b/infra/app-rails/service/main.tf
@@ -132,6 +132,5 @@ module "service" {
   is_temporary = local.is_temporary
 
   # Template Divergent Variables
-  container_read_only = true
-  healthcheck_type    = "curl"
+  healthcheck_type = "curl"
 }
--- a/infra/modules/service/main.tf
+++ b/infra/modules/service/main.tf
@@ -90,7 +90,7 @@ resource "aws_ecs_task_definition" "app" {
       cpu                    = var.cpu,
       networkMode            = "awsvpc",
       essential              = true,
-      readonlyRootFilesystem = var.container_read_only && !var.enable_command_execution,
+      readonlyRootFilesystem = !var.enable_command_execution,
 
       # Need to define all parameters in the healthCheck block even if we want
       # to use AWS's defaults, otherwise the terraform plan will show a diff
--- a/infra/modules/service/main.tf
+++ b/infra/modules/service/main.tf
@@ -31,6 +31,19 @@ locals {
       { name : name, value : value }
     ],
   )
+
+  ephemeral_write_volumes_with_name = [for container_path in var.ephemeral_write_volumes : {
+    container_path : container_path,
+    volume_name : replace(trim(container_path, "/"), "/", "_"),
+  }]
+  ephemeral_write_volumes = [for e in local.ephemeral_write_volumes_with_name : {
+    mount_point : {
+      "sourceVolume" : e.volume_name
+      "containerPath" : e.container_path,
+      "readOnly" : false
+    },
+    volume : { "name" : e.volume_name }
+  }]
 }
 
 #-------------------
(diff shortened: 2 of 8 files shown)


infra/modules/service/variables.tf
```diff
@@ -170,13 +170,13 @@ variable "vpc_id" {
...
 }
 
-# Custom Template-diverging variables
-variable "container_read_only" {
-  type        = bool
-  description = "Whether the container root filesystem should be read-only"
-  default     = true
+variable "ephemeral_write_volumes" {
+  type        = set(string)
+  description = "A set of absolute paths in the container to be mounted as writable for the life of the task. These should also be declared with VOLUME statements in the container definition file."
```
lorenyu, line 175: non-blocking idea, since I wonder if container definition could be confused by container-definition.json in AWS which is a separate thing

```suggestion
  description = "A set of absolute paths in the container to be mounted as writable for the life of the task. These should also be declared with VOLUME statements in the container definition file (Dockerfile)."
```

infra/modules/service/variables.tf
```diff
@@ -170,13 +170,13 @@ variable "vpc_id" {
...
+  default     = []
 }
 
+# Custom Template-diverging variables
```
lorenyu, line 179: Question unrelated to this PR, but Are these needed for template-application-rails to work? Should we get rid of these template-divergent variables if they aren't needed?

infra/modules/service/main.tf
```diff
@@ -113,13 +126,20 @@ resource "aws_ecs_task_definition" "app" {
           "awslogs-region"        = data.aws_region.current.name,
           "awslogs-stream-prefix" = local.log_stream_prefix
         }
-      }
-      mountPoints    = []
+      },
+      mountPoints    = [for e in local.ephemeral_write_volumes : e.mount_point]
```
lorenyu, line 130: style nit
```suggestion
      mountPoints    = [for volume in local.ephemeral_write_volumes : volume.mount_point]
```

or if we end up renaming local.ephemeral_write_volumes to local.ephemeral_write_volume_configs then do one of `[for config in ...]` or `[for volume_config in ...]`

infra/modules/service/main.tf
lorenyu, line 1: nit: `var.ephemeral_write_volumes` and `local.ephemeral_write_volumes` have the same name but different type/structure, one is a list of paths and one is a list of config objects. Since they are different types let's name them differently, e.g. do one of the following:

1. rename var.ephemeral_write_volumes to var.ephemeral_write_volume_paths
2. OR rename local.ephemeral_write_volumes to local.ephemeral_write_volume_configs
3. OR do both (1) and (2)

infra/modules/service/main.tf
```diff
@@ -113,13 +126,20 @@ resource "aws_ecs_task_definition" "app" {
...
       systemControls = []
       volumesFrom    = []
     }
   ])
 
+  dynamic "volume" {
+    for_each = [for e in local.ephemeral_write_volumes : e.volume]
+    content {
+      name = volume.value["name"]
```
lorenyu, line 139:
```suggestion
    for_each = [for volume_config in local.ephemeral_write_volume_configs : volume_config.volume]
    content {
      name = volume_config.value["name"]
```

infra/app-rails/service/main.tf
```diff
@@ -125,9 +125,12 @@ module "service" {
     } : {},
   )
 
+  ephemeral_write_volumes = [
+    "/rails/tmp"
+  ]
```
lorenyu, line 130: can we move this to service config (currently in app-config/env-config/outputs.tf even though we should probably move it to it's own file app-config/env-config/service.tf like we do with other config objects) so that we can minimize the number of customizations to the root modules

```suggestion
  ephemeral_write_volumes = local.service_config.ephemeral_write_volumes
```

lorenyu (approved):
Looks good just a couple of suggestions:
1. Move this new feature to be customizable in app-config
2. Rename at least one of var.ephemeral_write_volumes or local.ephemeral_write_volumes within modules/service to avoid confusion

infra/modules/service/main.tf
doshitan: For [dynamic blocks](https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks) the `for_each` var doesn't determine the iterator name, but can make it a little more compact with:

```
  dynamic "volume" {
    for_each = local.ephemeral_write_volume_configs[*].volume
    content {
      name = volume.value["name"]
    }
  }
```

As I don't think this is any clearer:
```
  dynamic "volume" {
    for_each = [for config in local.ephemeral_write_volume_configs : config.volume]
    iterator = "config"
    content {
      name = config.value["name"]
    }
  }
```
And we don't currently use `iterator` in our other dynamic blocks.

infra/modules/service/variables.tf
doshitan: > Are these needed for template-application-rails to work

Yes, as noted in 6274c142785a73a622852f9b9ed2df0c2f61102c (though the healthcheck bit/curl support is a hidden requirement not listed in upstream docs/not strictest of requires if we were to change the rails Dockerfile to include `wget`). But once this PR and https://github.com/navapbc/platform-test/pull/176 are merged everywhere, `template-application-rails` won't require infra edits (besides configuring `ephemeral_write_volumes`).

infra/modules/service/main.tf
lorenyu: Ah ok yeah the first code block looks good

doshitan:
Done in https://github.com/navapbc/template-application-rails/pull/73, https://github.com/navapbc/template-infra/pull/871, and https://github.com/navapbc/platform-test/commit/bb8ef195561136af29a1ce8dce4c3b9d281994b8

PR #176, opened 2025-02-07
Use service provided healthcheck, with fallbacks

doshitan:
## Changes

Rather than support a limited set of configurable healthcheck options, prefer a defined `healthcheck` executable that the service image provides (since it will best know what healthy is for itself), falling back through some common HTTP options if a `healthcheck` executable is not provided.

Longer term this also better supports folks trimming down their image size.

## Context for reviewers

Helps running `template-application-rails` apps without deviations from upstream `template-infra` .

- May want to choose a more specific name than `healthcheck` for the convention
- Not sure how many fallback options to realistically support, the bash one could be replaced by a similar netcat one if we think that'd cover more bases
- We'll want to document the convention of providing an `healthcheck` executable in `template-infra` docs

Breakdown:
- `app-rails/` changes not strictly required, but would go to `template-application-rails`
- `infra/modules/` changes go to `template-infra`
- `infra/app-rails/` changes apply to this repo (after previous points are merged)

## Testing

See PR Env deploys complete.

- Service endpoint: http://p-176-app-dev-1227445401.us-east-1.elb.amazonaws.com
- Deployed commit: fed685b5d42315c38b840489d2042a0e4750a937

- Service endpoint: http://p-176-app-rails-dev-1776651285.us-east-1.elb.amazonaws.com
- Deployed commit: fed685b5d42315c38b840489d2042a0e4750a937

### DIFF
--- a/infra/modules/service/main.tf
+++ b/infra/modules/service/main.tf
@@ -100,7 +100,7 @@ resource "aws_ecs_task_definition" "app" {
         retries  = 3,
         timeout  = 5,
         command = ["CMD-SHELL",
-          var.healthcheck_type == "curl" ? "curl --fail http://localhost:${var.container_port}/health || exit 1" : "wget --no-verbose --tries=1 --spider http://localhost:${var.container_port}/health || exit 1"
+          "([ -x \"$(command -v healthcheck)\" ] && healthcheck) || ([ -x \"$(command -v wget)\" ] && wget -q --spider http://127.0.0.1:$PORT/health) || ([ -x \"$(command -v curl)\" ] && curl --fail --silent http://localhost:$PORT/health > /dev/null) || ([ -x \"$(command -v bash)\" ] && bash -c \"exec 3<>/dev/tcp/127.0.0.1/$PORT;echo -e 'GET /health HTTP/1.1\\r\\nHost: http://localhost\\r\\nConnection: close\\r\\n\\r\\n' >&3;grep -q '^HTTP/1.1 200 OK' <&3\") || exit 1"
         ]
       },
       environment = local.environment_variables,
--- a/app-rails/Dockerfile
+++ b/app-rails/Dockerfile
@@ -117,7 +117,6 @@ ENV RAILS_ENV="production" \
 # Install packages needed for deployment
 RUN apt-get update -qq \
     && apt-get install -y --no-install-recommends \
-        curl \
         libvips \
         libyaml-dev \
         postgresql-client \
--- a/infra/app-rails/service/main.tf
+++ b/infra/app-rails/service/main.tf
@@ -130,7 +130,4 @@ module "service" {
   ]
 
   is_temporary = local.is_temporary
-
-  # Template Divergent Variables
-  healthcheck_type = "curl"
 }
--- a/infra/modules/service/variables.tf
+++ b/infra/modules/service/variables.tf
@@ -175,14 +175,3 @@ variable "ephemeral_write_volumes" {
   description = "A set of absolute paths in the container to be mounted as writable for the life of the task. These should also be declared with VOLUME statements in the container definition file."
   default     = []
 }
-
-# Custom Template-diverging variables
-variable "healthcheck_type" {
-  type        = string
-  description = "Whether to configure a curl or wget healthcheck. curl is more common. use wget for alpine-based images"
-  default     = "wget"
-  validation {
-    condition     = contains(["curl", "wget"], var.healthcheck_type)
-    error_message = "choose either: curl or wget"
-  }
-}


infra/modules/service/main.tf
```diff
@@ -100,7 +100,7 @@ resource "aws_ecs_task_definition" "app" {
         retries  = 3,
         timeout  = 5,
         command = ["CMD-SHELL",
-          var.healthcheck_type == "curl" ? "curl --fail http://localhost:${var.container_port}/health || exit 1" : "wget --no-verbose --tries=1 --spider http://localhost:${var.container_port}/health || exit 1"
+          "([ -x \"$(command -v healthcheck)\" ] && healthcheck) || ([ -x \"$(command -v wget)\" ] && wget -q --spider http://127.0.0.1:$PORT/health) || ([ -x \"$(command -v curl)\" ] && curl --fail --silent http://localhost:$PORT/health > /dev/null) || ([ -x \"$(command -v bash)\" ] && bash -c \"exec 3<>/dev/tcp/127.0.0.1/$PORT;echo -e 'GET /health HTTP/1.1\\r\\nHost: http://localhost\\r\\nConnection: close\\r\\n\\r\\n' >&3;grep -q '^HTTP/1.1 200 OK' <&3\") || exit 1"
```
lorenyu, line 103: I have a preference not to include the `raw TCP request version using bash` for now until/unless a project needs it. It's nice but prefer to limit unused code. Furthermore it makes the line harder to read, and technically makes removing it in the future a breaking change since we can't guarantee no project is relying on that behavior.

Longer term I even lean slightly towards not having any fallbacks to keep the simpler e.g. `command = ["CMD-SHELL", "healthcheck || exit 1"]`. It's less code and doesn't reduce an application's team flexibility. It's also less magic and it simplifies the contract with applications since the `healthcheck` executable requirement also effectively replaces the need to have a `/health` endpoint. The contract becomes "the container needs a `healthcheck` executable", rather than "the container needs a `healthcheck` executable OR it needs `wget` and a `/health` endpoint).

lorenyu (approved):

infra/modules/service/main.tf
doshitan: > I have a preference not to include the `raw TCP request version using bash`
> for now until/unless a project needs it. It's nice but prefer to limit unused
> code.

If we proceed with all the changes in this PR, `curl` will be the unused bit, as
only the Rails template uses it and these changes drop installing it in
production image since `bash` is included already there.

And kinda similar for all of the other base images our templates use, except for
`template-infra/template-only-app`. Their base images all include Bash (+grep)
and we install `wget` specially for the healthcheck, which we could start
dropping. `template-infra/template-only-app` is a little different in that it is
based on Alpine, and so does not include Bash in the base image but does
include `wget` (and `nc`).

Bash is the biggest of all (~6MB), but already present in most of our images and
more likely to be utilized by utility scripts in production.

> Furthermore it makes the line harder to read

Yea, it's a little long to embed directly and the double escaping is annoying.

> technically makes removing it in the future a breaking change since we can't
> guarantee no project is relying on that behavior.

How's it different than any of the other ones? Or do you mean, it'd kinda be an
_additional_ breaking change to removing wget/curl that we could avoid by not
supporting in the first place?

> Longer term I even lean slightly towards not having any fallbacks to keep the
> simpler e.g. `command = ["CMD-SHELL", "healthcheck || exit 1"]`. It's less
> code and doesn't reduce an application's team flexibility. It's also less
> magic and it simplifies the contract with applications since the `healthcheck`
> executable requirement also effectively replaces the need to have a `/health`
> endpoint. The contract becomes "the container needs a `healthcheck`
> executable", rather than "the container needs a `healthcheck` executable OR it
> needs `wget` and a `/health` endpoint.

Agreed. Though I paused pushing that direction too hard right now because the
`infra/modules/service` is so heavily geared towards a single web app exposed
via a single port with a `/health` endpoint (with a few other things dependent
on that as well at the moment), so relatively simple to provide some basic
support out of the box for our users.

lorenyu: > How's it different than any of the other ones? Or do you mean, it'd kinda be an
additional breaking change to removing wget/curl that we could avoid by not
supporting in the first place?

Yeah exactly.

---

Thanks for the helpful context, all the points make sense to me, as long as we're aligned on the goals (reducing unused code and avoiding adding new features that aren't strictly needed and that we might remove later) I'm fine with your judgement on which of these to keep or remove.

From the perspective of minimizing image size, it sounds like maybe the only ones we need are actually the wget and bash versions, but maybe from the perspective of simplicity to read and understand the wget and curl ones are a bit easier to parse. I don't feel strongly about it, and if you decide to keep all three that'd be fine too.

Maybe one suggestion would be to add a comment block in front of the healthcheck command line explaining all of the commands, since it's not that easy to read since it has to be on a single line. Maybe something like:

```
# If there is a `healthcheck` executable in the container $PATH then use that, otherwise fall back to one of: wget, curl, and raw TCP request using bash
```

PR #182, opened 2025-02-24
Generate infra service test id from timestamp

lorenyu:
## Ticket

Work towards https://github.com/navapbc/template-infra/issues/787

## Changes

see title

## Context for reviewers

This is the first step towards https://github.com/navapbc/template-infra/issues/787

The ultimate goal is to have a way to look for stale environments generated by the infra service tests. The way we plan on doing that is to encode timestamp information in the id itself, but because we want the id to be short we are base62 encoding the id.

## Testing

I looked at the ci-infra-app-service.yml and ci-infra-app-rails-service.yml workflows and manually decoded the base62 ids, then compared the resulting timestamps to the UTC timestamps that the jobs started on and saw that they roughly matched.

- Service endpoint: http://p-182-app-dev-8425939.us-east-1.elb.amazonaws.com
- Deployed commit: a250e576473d7aa7e9240d102a0cab02a181ac83

- Service endpoint: http://p-182-app-rails-dev-1852731477.us-east-1.elb.amazonaws.com
- Deployed commit: a250e576473d7aa7e9240d102a0cab02a181ac83

### DIFF
--- a/infra/test/infra_test.go
+++ b/infra/test/infra_test.go
@@ -16,12 +16,11 @@ import (
 	"time"
 
 	http_helper "github.com/gruntwork-io/terratest/modules/http-helper"
-	"github.com/gruntwork-io/terratest/modules/random"
 	"github.com/gruntwork-io/terratest/modules/shell"
 	"github.com/gruntwork-io/terratest/modules/terraform"
 )
 
-var uniqueId = strings.ToLower(random.UniqueId())
+var uniqueId = strings.ToLower(generateTestId())
 var workspaceName = fmt.Sprintf("t-%s", uniqueId)
 var testAppName = os.Getenv("APP_NAME")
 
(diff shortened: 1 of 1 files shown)


infra/test/infra_test.go
```diff
@@ -94,3 +93,30 @@ func DestroyService(t *testing.T, terraformOptions *terraform.Options) {
...
 	fmt.Println("::endgroup::")
 }
+
+func generateTestId() string {
+	now := time.Now()
+	// Combine hour, minute, second, and millisecond into a single number
+	timeNum := int64(now.Hour())*3600000 +
+		int64(now.Minute())*60000 +
+		int64(now.Second())*1000 +
+		int64(now.Nanosecond()/1000000)
```
doshitan, line 103: Can we just use the normal timestamp, like:
```suggestion
	timeNum := now.Unix()
```

lorenyu: I do like that it's much much simpler. I think I'll go with it and just be less paranoid.

doshitan: Ah, right this doesn't run under PR workspace, creates it's own that we are generating the name for, that's the whole point. Could use `UnixMilli` (or `UnixMicro`) if it becomes an issue? Though that will add a character to the string length I think.

Two tests starting up at the exact same second _feels_ pretty unlikely, but I dunno.

lorenyu: Looks like UnixMilli base62 encoded is 8 characters, so t-<id> would be 10 characters. I believe there's a task to come up with char length budgets for each name but we haven't done that yet so not sure if this is too long. Happy to stick with seconds for now. I think collisions should be infrequent enough and a quick thought exercise makes me think it's recoverable.

lorenyu:
Done in https://github.com/navapbc/template-infra/pull/885

PR #184, opened 2025-02-28
Move test library files to lib

lorenyu:
## Ticket

## Changes

## Context for reviewers

It would be valuable to have useful utility classes like [the EmailService](https://github.com/navapbc/pfml-starter-kit-app/tree/main/e2e/services) that can be used for e2e tests across a variety of applications, but we should organize them in a folder rather than putting them all directly in /e2e.

## Testing

CI should run e2e tests


lorenyu:
Nvm let me do this PR in pfml-starter-kit-app

PR #193, opened 2025-03-14
⚠️ Remove create user roles migration

lorenyu:
## Ticket

n/a

## Changes

see title

## Context for reviewers

This commit https://github.com/navapbc/template-application-rails/commit/5e2a120c1999c5b3632aac800526d6bfc862a734 removes the user roles from the application code but not from the database. This PR removes it from migrations. This was separated out since removing an already existing migration is not backwards compatible. Since this is a template it makes sense to do it this way rather than add an extra migration that removes the user_roles table.

## Migration notes

When updating the template in this version, revert the changes in these files:
- <APP_NAME>/db/migrate/20240410213056_create_user_roles.rb (deleted in this change)
- <APP_NAME>/db/schema.rb (user_tables removed)

## Testing

Removed the "create_user_roles" migration file and regenerated schema.db by running `make db-reset`, `make clean-volumes`, and `make init-db` in app-rails app

- Service endpoint: https://p-193-app-dev-1347276088.us-east-1.elb.amazonaws.com
- Deployed commit: daeef1587462521c83a5b44582812902e927aca8

- Service endpoint: https://p-193-app-rails-dev-1614161222.us-east-1.elb.amazonaws.com
- Deployed commit: daeef1587462521c83a5b44582812902e927aca8


lorenyu:
Done in https://github.com/navapbc/template-application-rails/pull/83

PR #196, opened 2025-04-15
[draft] Update CI vulnerability workflow to reduce how often the docker image is built

lisac:
## Ticket

## Changes

Refactors docker image creation in the `vulnerability-scans.yml` workflow:
- new job: `build-and-cache`: builds the image and puts it in a cache
- updates other jobs to use the cached image, instead of building the image as part of the job

## Context for reviewers

Based off of the implementation in HHS/simpler-grants-gov:  https://github.com/HHS/simpler-grants-gov/blob/6dadc31f8474c386b824654d3ce2bb428bec584f/.github/workflows/vulnerability-scans.yml 

## Testing

### DIFF
--- a/.github/workflows/vulnerability-scans.yml
+++ b/.github/workflows/vulnerability-scans.yml
@@ -85,12 +85,6 @@ jobs:
           --cache-from=type=local,src=/tmp/.buildx-cache \
           --cache-to=type=local,dest=/tmp/.buildx-cache"
 
-      - name: Check disk space, including docker usage
-        if: success() || failure()
-        run: |
-          df -h
-          docker system df
-
       - name: Save Docker image
         if: steps.cache-buildx.outputs.cache-hit != 'true'
         run: |
@@ -102,12 +96,6 @@ jobs:
         with:
           path: /tmp/docker-image.tar
           key: ${{ inputs.app_name }}-docker-image-${{ github.sha }}
-
-      - name: Check disk space, including docker usage
-        if: success() || failure()
-        run: |
-          df -h
-          docker system df
 
   trivy-scan:
     runs-on: ubuntu-latest
--- a/.github/workflows/vulnerability-scans.yml
+++ b/.github/workflows/vulnerability-scans.yml
@@ -112,6 +112,7 @@ jobs:
 
   trivy-scan:
     runs-on: ubuntu-latest
+    needs: build-and-cache
 
     steps:
       - uses: actions/checkout@v4
@@ -128,19 +129,23 @@ jobs:
         with:
           files: ${{ inputs.app_name }}/trivy-secret.yaml trivy-secret.yaml
 
-      - name: Build and tag Docker image for scanning
-        id: build-image
+      - name: Restore cached Docker image
+        uses: actions/cache/restore@v4
+        with:
+          path: /tmp/docker-image.tar
+          key: ${{ inputs.app_name }}-docker-image-${{ github.sha }}
+          restore-keys: |
+            ${{ inputs.app_name }}-docker-image-
+
+      - name: Load cached Docker image
         run: |
-          make APP_NAME=${{ inputs.app_name }} release-build
-          IMAGE_NAME=$(make APP_NAME=${{ inputs.app_name }} release-image-name)
-          IMAGE_TAG=$(make release-image-tag)
-          echo "image=$IMAGE_NAME:$IMAGE_TAG" >> "$GITHUB_OUTPUT"
+          docker load < /tmp/docker-image.tar
 
       - name: Run Trivy vulnerability scan
         uses: aquasecurity/trivy-action@master
         with:
           scan-type: image
-          image-ref: ${{ steps.build-image.outputs.image }}
+          image-ref: ${{ needs.build-and-cache.outputs.image }}
           format: table
           exit-code: 1
           ignore-unfixed: true
--- a/.github/workflows/vulnerability-scans.yml
+++ b/.github/workflows/vulnerability-scans.yml
@@ -200,6 +200,7 @@ jobs:
 
   dockle-scan:
     runs-on: ubuntu-latest
+    needs: build-and-cache
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


lisac:
@lorenyu: your notes here make sense to me. Thanks for taking the time to walk through this. 
I'm not confident about having revisions reviewable this week - next week more likely - will re-request a review when I have those changes in. Thanks again!

lisac:
status: making progress, ran into what i think is a fixable bug and will revisit next week.

- good: moved the build logic into a composite action used by `build-and-publish` and `vulnerability-scans`; observed desired `concurrency` behavior, to avoid multiple workflows executing the build action concurrently
- bug: the cache lookup is not behaving as i expected. observed: the two workflows (`build-and-publish`  and `vulnerability-scans`) apparently used different SHAs in their cache keys, whereas i expected the cache keys to be identical.  they keys they used:
  - `platform-test-app:1a6cc698c726109c799bfca714a2a48496e76cd3`
  - `platform-test-app:ba457e9ca8cc127a1361b1fc72e4cf56f9c6105f`

next steps:
- fix the bug and remove debugging 
- update the Issue's description
- re-request PR reviews

.github/workflows/build-and-publish.yml
```diff
@@ -79,10 +81,48 @@ jobs:
           echo "Is image published: $is_image_published"
           echo "is_image_published=$is_image_published" >> "$GITHUB_OUTPUT"
 
-      - name: Build release
```
lorenyu, line 82: what's the reason for making this a separate job?

.github/workflows/build-and-publish.yml
```diff
@@ -48,16 +48,18 @@ jobs:
           echo "Commit hash: $COMMIT_HASH"
           echo "commit_hash=$COMMIT_HASH" >> "$GITHUB_OUTPUT"
 
-  build-and-publish:
-    name: Build and publish
+  check-image-already-published:
+    name: Check whether the image is already published
     runs-on: ubuntu-latest
     needs: get-commit-hash
-    concurrency: build-and-publish-${{ inputs.app_name }}-${{ needs.get-commit-hash.outputs.commit_hash }}
```
lorenyu, line 55: this concurrency is important to keep. the concurrency statement later on that uses `github.ref` won't work since different github.refs can refer to the same commit hash (e.g. main, origin/main, HEAD, <commit hash>, <version tag>) can all be valid refs that point to the same commit hash, so there'd be a race condition where multiple jobs are trying to build the same commit hash but don't realize it since they are referencing the commit via different refs. that's why we have the separate job beforehand that gets the commit hash.

.github/workflows/build-and-publish.yml
```diff
@@ -79,10 +81,48 @@ jobs:
...
+      - name: Restore cached Docker image
+        uses: actions/cache/restore@v4
+        with:
+          path: /tmp/docker-image.tar
+          key: ${{ steps.build-release-candidate.outputs.image_cache_key }}
+          fail-on-cache-miss: true
+
+      - name: Load cached Docker image
+        run: |
+          docker load < /tmp/docker-image.tar
```
lorenyu, line 116: can we put these steps into actions/build-release-candidate

.github/workflows/ci-app-rails-vulnerability-scans.yml
```diff
@@ -26,3 +26,4 @@ jobs:
     uses: ./.github/workflows/vulnerability-scans.yml
     with:
       app_name: "app-rails"
+      ref: ${{ github.ref }}
```
lorenyu, line 29: is this needed?

.github/actions/build-release-candidate/action.yml
```diff
@@ -0,0 +1,52 @@
...
+        echo "image=$IMAGE_NAME:$IMAGE_TAG" >> "$GITHUB_OUTPUT"
+      shell: bash
+
+    - name: Check whether the image has already been built
+      id: check-image-already-exists
+      uses: actions/cache@v4
+      with:
+        path: /tmp/docker-image.tar
+        key: ${{ steps.create-image-identifier.outputs.image }}
+        lookup-only: true
```
lorenyu, line 29: what's the reason for only doing a lookup? i think we'd want to download the image if it's already built

.github/actions/build-release-candidate/action.yml
```diff
@@ -0,0 +1,52 @@
...
+      run: |
+        docker save ${{ steps.create-image-identifier.outputs.image }} > /tmp/docker-image.tar
+      shell: bash
+
+    - name: Cache Docker image
+      if: steps.check-image-already-exists.outputs.cache-hit != 'true'
+      uses: actions/cache/save@v4
+      with:
+        path: /tmp/docker-image.tar
+        key: ${{ steps.create-image-identifier.outputs.image }}
```
lorenyu, line 48: this step shouldn't be needed. cache should save automatically when the job completes

.github/actions/build-release-candidate/action.yml
```diff
@@ -0,0 +1,52 @@
...
+
+    - name: Check whether the image has already been built
+      id: check-image-already-exists
+      uses: actions/cache@v4
+      with:
+        path: /tmp/docker-image.tar
+        key: ${{ steps.create-image-identifier.outputs.image }}
+        lookup-only: true
+
+    - name: Build and tag Docker image for scanning
```
lorenyu, line 31: Not necessarily in scope for this PR if it is getting complex, but in an older version of this PR you had code that also cached the intermediate docker layers in `/tmp/.buildx-cache` which could dramatically speed up builds even when it's a cache miss since some of the intermediate layers will be cached

lorenyu (commented):
I realize you're still working on this, but i saw you left a comment so i decided to poke and left some comments. feel free to ignore if they are a distraction but hopefully they help

.github/actions/build-release-candidate/action.yml
lisac: Yes, you're right. I removed that from my branch because I was overwhelmed by all the other issues I was running into. At this point - I want to get the PR closed! - I think I won’t attempt to get that feature working. Should I propose it as a new issue? or add it as a comment to https://github.com/navapbc/template-infra/issues/206 ?

lisac:
@lorenyu : I appreciate these notes - thank you. For what you’ve flagged, most of those were knowledge gaps on my part, or me thinking too narrowly around the vulnerability scans workflow. I’ll be following your notes to get this PR cleaned up.

.github/actions/build-release-candidate/action.yml
lorenyu: Sure go ahead and create a new issue and link it here

lorenyu:
@lisac sounds good! happy to help if you run into any issues

.github/actions/build-release-candidate/action.yml
lisac: https://github.com/navapbc/template-infra/issues/936

PR #203, opened 2025-04-25
Enable local login for dev env

LeslieJAnderson:
## Ticket

Resolves https://github.com/navapbc/template-application-rails/issues/55

## Changes

> Adds simple local authentication workaround

## Context for reviewers

>Unblocks development if the developer is without proper creds yet

## Testing

> Logged in repeatedly

- Service endpoint: https://p-203-app-dev-2096639689.us-east-1.elb.amazonaws.com
- Deployed commit: 0b49f28c540088351b081ce46705dcd7b726450e

- Service endpoint: https://p-203-app-rails-dev-713786069.us-east-1.elb.amazonaws.com
- Deployed commit: 0b49f28c540088351b081ce46705dcd7b726450e


app-rails/app/services/auth_service.rb
```diff
@@ -1,8 +1,18 @@
...
 
 class AuthService
-  def initialize(auth_adapter = Auth::CognitoAdapter.new)
-    @auth_adapter = auth_adapter
+  def initialize(auth_adapter = nil)
+    @auth_adapter = auth_adapter || default_adapter
+  end
+
+  def default_adapter
+    return Auth::MockAdapter.new if Rails.env.test?
```
lorenyu, line 9: i didn't realize we already have a MockAdapter, can we just use this one instead of a separate LocalAdapter?

app-rails/app/services/auth_service.rb
```diff
@@ -1,8 +1,18 @@
...
+
+    if Rails.env.development? && ENV["COGNITO_CLIENT_SECRET"].to_s.empty?
```
lorenyu, line 11: can we add a new environment variable instead like `AUTH_SERVICE` which can have values `"cognito"` or `"local"` or `"mock"` (or just "cognito" and "mock" depending on the answer to my previous question about reusing mock)?

lorenyu (commented):
Awesome thanks! I love how simple this is. I had a question about whether we can make it even simpler by consolidating the mock adapter with the local adapter into one. It feels like they are both pretty simple and we could combine somehow.

app-rails/app/services/auth_service.rb
LeslieJAnderson: Good catch — while MockAdapter simulates the auth flow for test-like cases and handles expected errors and challenges, LocalAdapter was meant as a local-only, in-memory login helper for devs, including token generation and dynamic user creation without string pattern checks.

I approached it this way to keep concerns separate: MockAdapter acts as a test double for external services, while LocalAdapter provides a lightweight, dev-only shortcut for local workflows without coupling to the full mock behavior. We could refactor and cnsolidate them, but I’d lean toward keeping them distinct to preserve that separation of concerns unless we start seeing overlap in how they're used.

lorenyu: I'd prefer to consolidate them with the lowest common denominator of what is needed. Local development arguably needs less than the test suite does, since devs just need a way to bypass authentication to get to the application. (If devs are working on authentication itself then they should be working with cognito directly not a local adapter or a mock)

I think the simplest behavior is along the lines of:
- auth challenge always succeeds except for specific passwords that trigger failures (used for testing error conditions like incorrect password)

Any other cases you think are needed?

LeslieJAnderson: Okay, I think I tackled this.

lorenyu (commented):
Left some comments. Main thing is how the config/env vars is set

app-rails/README.md
```diff
@@ -96,6 +96,15 @@ To run natively:
 1. `make start-native`
 1. Then visit http://localhost:3100
 
+#### Local Authentication
+
+You can circumvent typical login and log in using any email and password if you add
+```
+AUTH_ADAPTER=mock
```
lorenyu, line 103: Since the default for local is already mock, maybe flip the instruction here to explain how to set to cognito (and explain that you also need to set the other cognito env vars)

lorenyu (approved):
Thanks! I think this is ready to move over to template-application-rails to merge from there, then we can close this PR since once we merge there it will automatically propagate the changes to platform-test

lorenyu:
Done in https://github.com/navapbc/template-application-rails/pull/101

PR #205, opened 2025-04-30
Lorenyu/lookbook

lorenyu:
## Ticket

## Changes

## Context for reviewers

## Testing

### DIFF
--- a/app-rails/app/previews/breadcrumbs_preview.rb
+++ b/app-rails/app/previews/breadcrumbs_preview.rb
@@ -1,6 +1,8 @@
 class BreadcrumbsPreview < Lookbook::Preview
+  layout "component_preview"
+
   def default
-    result = render template: "application/_breadcrumbs", locals: { crumbs: [
+    render template: "application/_breadcrumbs", locals: { crumbs: [
       { name: "Passport applications", url: "https://google.com" },
       ], current_name: "New passport application" } 
   end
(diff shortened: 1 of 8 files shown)


app-rails/app/previews/breadcrumbs_preview.rb
```diff
@@ -0,0 +1,9 @@
+class BreadcrumbsPreview < Lookbook::Preview
+  layout "component_preview"
+
+  def default
+    render template: "application/_breadcrumbs", locals: { crumbs: [
```
lorenyu, line 5: `render partial: "application/breadcrumbs"` not supported by lookbook so use `render template:` and remember to prefix with _ and set layout to "component_preview" to get rid of application layout

app-rails/app/views/application/_header.html.erb
```diff
@@ -29,12 +29,12 @@
             <% end %>
 
           <li class="usa-nav__primary-item">
-            <%= render partial: "language-toggle", locals: { container_class: "display-block desktop:display-none relative" } %>
+            <%= render partial: "application/language-toggle", locals: { container_class: "display-block desktop:display-none relative" } %>
```
lorenyu, line 32: We need to explicitly set partial paths rather than rely on Rails fallback behavior. This is probably better practice anyways, easier to read/find things for the reader

PR #208, opened 2025-06-05
Fix PR environment destroy race condition

lorenyu:
## Ticket

Resolves https://github.com/navapbc/template-infra/issues/937

## Changes

see title 

## Context for reviewers

I noticed that occasionally the PR environment gets destroyed (successfully) before it gets created again. Looking into it, it looks like if the PR is closed or merged before the PR environment check's build-and-publish job is completed, the PR environment destroy job will actually run first. This change fixes that issue by moving the concurrency key to operate at the workflow level rather than just at on the job level.

Or more precisely:
1. Dev pushes some final changes to PR before merge
2. PR env update triggers, starts building container to deploy
3. Dev merges PR
4. PR env delete triggers
5. PR env update finishes building, hits concurrency lock on doing the deploy
6. PR env delete completes
7. PR env update is unblocked and deploys updated code, which actually recreates the entire env

## Testing

Ran PR environment destroy soon after PR environment checks started running, and made sure PR environment was actually destroyed

image

image

image

---

♻️ Environment destroyed ♻️

- Service endpoint: https://p-208-app-rails-dev-555674829.us-east-1.elb.amazonaws.com
- Deployed commit: cf508e7b7ddd4e04667d1dfbaa141262b1dfb40f

### DIFF
--- a/.github/workflows/pr-environment-checks.yml
+++ b/.github/workflows/pr-environment-checks.yml
@@ -15,6 +15,9 @@ on:
       commit_hash:
         required: true
         type: string
+
+concurrency: pr-environment-${{ inputs.app_name }}-${{ inputs.pr_number }}
+
 jobs:
   build-and-publish:
     name: " " # GitHub UI is noisy when calling reusable workflows, so use whitespace for name to reduce noise
@@ -33,8 +36,6 @@ jobs:
       id-token: write
       pull-requests: write # Needed to comment on PR
       repository-projects: read # Workaround for GitHub CLI bug https://github.com/cli/cli/issues/6274
-
-    concurrency: pr-environment-${{ inputs.app_name }}-${{ inputs.pr_number }}
 
     outputs:
       service_endpoint: ${{ steps.update-environment.outputs.service_endpoint }}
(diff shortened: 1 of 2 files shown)


MichaelCrawfordNava (approved):
LGTM!

doshitan (approved):
So the situation is:
1. Dev pushes some final changes to PR before merge
2. PR env update triggers, starts building container to deploy
3. Dev merges PR
4. PR env delete triggers
5. PR env update finishes building, hits concurrency lock on doing the deploy
6. PR env delete completes
7. PR env update is unblocked and deploys updated code, which actually recreates the entire env

Regardless, I think it makes sense that workflows involved with PR environments shouldn't be running concurrently.

As a future followup, I wonder if the PR env delete workflow should also set an appropriate `cancel-in-progress`, so we don't burn CI minutes waiting for an environment to finish updating that is just going to be immediately destroyed. That could result in some strange terraform state issues though.

lorenyu:
> That could result in some strange terraform state issues though.

Yeah I was mainly worried about how aggressive cancel-in-progress is, like will it at least release the terraform state lock from DynamoDB? But if it's just like Ctrl+C then I think that's a good idea.

lorenyu:
Done in https://github.com/navapbc/template-infra/pull/945

PR #228, opened 2025-10-03
Replace current-region script with aws configure get region

lorenyu:
## Ticket

## Changes

- Replace calls to current-region with aws configure get region
- Remove current-region script

## Context for reviewers

Recently there's been sporadic errors where the current-region script would return `:` instead of the region. This approach should be more robust and also remove the need for the extra current-region script.

## Testing

CI should cover at least a few of the changes (e.g. in publish-release)

### DIFF
--- a/bin/current-region
+++ b/bin/current-region
@@ -1,4 +1,0 @@
-#!/bin/bash
-# Print the current AWS region
-set -euo pipefail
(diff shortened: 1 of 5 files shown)


lorenyu:
Doesn't seem to work

